PR #31, opened 2016-04-27
Added package.json for installation with npm

fabiantheblind:
We (@fabiantheblind & @vogelino) are trying to create a build process for Extendscript that allows us to use es6 in Extendscript. To be able clone basil.js with npm we need a package.json file.  

The installation would then be:

    npm install git+ssh:git@github.com/basiljs/basil.js.git
    # or
    npm install git+https:git@github.com/basiljs/basil.js.git

Wouldn't that be awesome?  
I bumped the patch version to 1.0.10 for that matter within the package.json and in basil.js


vogelino:
Indeed that would be really great! The main problem from now is the way basil includes other files with the {{#include}} keyword. It would be better if the whole scripts were bundled in one single file which could be imported as a npm module.

looking forward to add it to the build process. Follow the progress at https://github.com/vogelino/es2015-2-extendScript-build

ludzeller:
Hello Fabian, hello Lucas,

Thanks for sending this, we appreciate your time and work! Unfortunately I am quite busy with non-basil.js things at the moment, so if we include this in the trunk, then it needs to be stable and tested from the beginning on...

Regarding the #include statements I can say that we originally decided to split the big file into smaller ones to make it easier to edit. Can't remember another reason right now. @b-g ?

Btw is npm actually free in for scenario? We wouldn't really need it for teaching, but if it makes it more attractive to use it, then it's of course interesting.

We could still offer the traditional install easily, right?
Any other changes?

fabiantheblind:
Hi @ludzeller npm is free and the normal install still works. We just need the package.json to clone the repo. There are no other changes. So there is nothing to test. It is just a file that has to be present. The changes for compiled/bundled version is another thing. We will currently use my fork with the bundle branch for that. see https://github.com/fabiantheblind/basil.js/tree/bundle

greenish:
This PR is not affecting any functionality of basil.js so I wouldn't see any reason to not merge it.

The only thing that should be considered is the `name` in `package.json`. Both `basil` and `basil.js` are taken on http://npmjs.com. `basiljs` however is free. 

@ludzeller I would recommend choosing a free name and publishing the package to npm.

Cheers

fabiantheblind:
Oh I didn't look if the names are taken. Damn. I'll make a PR today with the changes.

b-g:
Hi @fabiantheblind & @vogelino,
Hi all,

Many thanks for the PR!! Good call! 

But before I merged it ... a few more mini details:
1. I will document the new versioning system in the changelog and add a tag to the PR/commit ... then it can be used in npm with a direct reference. 
2. @fabiantheblind I was wondering whether we really want to publish the entire repo, or if we want to keep it a bit leaner an exclude a few things via a [.npmignore](https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package)? Thoughts?
3. @ludzeller / @ffd8 someone of you guys (how is control of github.com/basiljs) should [publish the package to npm](https://docs.npmjs.com/getting-started/publishing-npm-packages) under a npm user with the same name. OK?

RE: "es6 in Extendscript" would be super! And I would love to refactor the basil core code to require/browserify style. However I heard from @andreaskoller (who tried something similar = Extendscript in After FX) that in the end the combo es6 to Extendscript did't work out for him. Andreas settled down to Typescript as this compiles out of the box to es5 (= compatible to Extendscript After FX). Maybe you guys should talk? :)

Best, 
B

b-g:
Merged + Many thanks! :) I'll wait for feedback on .npmignore before I add the tag ...

fabiantheblind:
I think it's not a lot to exclude maybe these files:

- basil.sublime-project
- install.bat
- install.command
- extras/Sublime Text

everything else could stay as is for now (IMHO).  
We can exclude things later on as the package evolves. The npm approach still is for pro users. They should know what to do with all those files.  

BTW. Sorry for introducing semver without mentioning it.

About the compilation from es6:

I already had it working in AE but now I get an error. 😠 I'm trying to pinpoint it. Andreas told me that he is using TypeScript but we would love to stay with JavaScript for now. If we come up with a working process we will report back.  

I already have a script (in python) that bundles Basil into a single file but this is also for another PR.

b-g:
Hi Fabian,

Ok, then I won't include a .npmignore ... seems to be an overkill for now.

RE: es6 compilation
Please keep us posted on es6 compilation to Extendscript! Would be really great to finally be able to code Extendscript in a nice way :)

RE: bundling Basil into a single file
Great! PR very welcome! However I think I would prefer then a Gulp or Grunt task for this, so that we stay in js realm (especially now that we finally have a package.json file)

Thanks! + Have a nice weekend!

b-g:
FYI
Latest commit tagged **1.0.10**
https://github.com/basiljs/basil.js/releases/tag/1.0.10

fabiantheblind:
Yes. I will report back and the build process should of course be in js. 
Cheers
Fabian

ffd8:
As I prepare the bundle for the basiljs.ch website... I think an extra 0 got introduced as the packages went from 1.08 -> 1.09 -> now 1.0.10 ?

Seems like this version should simply be 1.10?

fabiantheblind:
I would suggest using semantic versioning. It is a standard for packages on npm. https://docs.npmjs.com/getting-started/semantic-versioning 


>Bug fixes and other minor changes: Patch release, increment the last number, e.g. 1.0.1  

>New features which don't break existing features: Minor release, increment the middle number, e.g. 1.1.0 

>Changes which break backwards compatibility: Major release, increment the first number, e.g. 2.0.0

It's a good system.

b-g:
Hi @ffd8, the additional "0" was added on purpose to make it compatible with the rest of the "js world". I hope this is okay for you. Best, B

ffd8:
Hey @b-g + @fabiantheblind , all is fine with me, was just surprised of the new 0 since it wasn't there in previous releases (or was lacking a . - ie. 1.0[.]8). Obviously I'm overly concerned on the vanity of the packaging... you know, for that pending basil.js unboxing video.

I totally agree, the npm js semantic versioning is a good way to go. Therefore wouldn't this release be 1.1.0? or still 1.0.10? .. eventually 1.0.11..? Even though it's a minor release.. doesn't 1.0.10 = 1.0.1 (decimal dropping at the end). It's not discussed on that their page Fabian referenced, but I'd guess that you shouldn't go above 9 on any decimal placement (only the major release can)?

*ahem.. then can I censor this conversation post decision.. haha probably embarrassing*

fabiantheblind:
Maybe this article is helpful https://nodesource.com/blog/semver-a-primer/

> So:
> - Version 0.3.10 is ordered before 0.10.3
> - Version 0.1.1 is ordered before 1.0.0
> - Version 1.100.100 is ordered before 10.10.10

ffd8:
Okaaay- haha, I guess I just have to think of it more like a IP address and less like a 3 part decimal number...

fabiantheblind:
Yeah. IP is a got "metaphor"

andyinabox:
FYI: I create very simple example for a project using basil.js via npm here: https://github.com/andyinabox/basiljs-template.

fabianmoronzirfas:
Hi @andyinabox thanks. Lets take this discussion over to your repo

PR #37, opened 2016-06-23
added roadmap for development

fabiantheblind:
Like we decided on our meeting. Here is the roadmap draft for things to come.

### DIFF
--- a/roadmap.txt
+++ b/roadmap.txt
@@ -1,0 +1,22 @@
+# Roadmap  
+
+Where is Basil.js heading? (This document should eventually be converted into milestones and issues)  
+
+## 2.0.0
+
+- [ ] get rid of the `b.` to make it more Processing like and even easier to write
+- [ ] bundle the library to make it a single file distribution  
+- [ ] ES6?  
+
+## Possible new features
+
+- [ ] HSL / HSB color mode  
+- [ ] `size()` function for setting page size  
+- [ ] set and get object properties easily  
+- [ ] add to story  
+- [ ] search for objects with certain properties (could be done with object search)  
+- [ ] set reference point  
+
+## Bugs  
+
+- [ ] canvas mode behaves not as described in some occasions (mentioned by @ffd8)


trych:
Hi Fabian,
unfortunately I missed the meeting (or rather, didn't know about it ;) ). But there is some good news, it just has been confirmed for me yesterday that I will teach a course on basil.js next semester at HAW Hamburg. I have a few ideas as well for the road map and a few features, that could be added.
Is there any chance to get in touch with you (or others) during the summer (maybe over Skype)? As I understand you have taught basil.js in the past (as have I, but only in a short introduction course) and I would like to hear about your experiences.

About the roadmap: I am not sure, if skipping the b. is actually an improvement. I think it is a nice feature to clearly distinguish basil.js functions (or rather methods) from "regular" Indesign scripting functions. Especially for beginners I think that would make things easier, if they come to a point where they mix regular InDesign scripting into their basil.js stuff. And typing b. is not really that much of a hassle, is it? ;)
 
Timo

fabiantheblind:
The meeting was some weeks ago between @b-g @ffd8 and me. We talked about what could go into v2 of Basil and agreed that removing the `b.` would be good thing. It is kind of cumbersome to always write the `b.`for longer scripts. The idea is to make it even more Processing like. Code taken from Processing could be easier ported. I gave a seminar this semester with Basil and the students always forgot to write the `b.`.   

A meeting would be great this summer and any help is also appreciated.

fabiantheblind:
> I think it is a nice feature to clearly distinguish basil.js functions (or rather methods) from "regular" Indesign scripting functions. 

BTW. There are no "real" InDesign scripting functions. Everything is bound to objects.

trych:
> BTW. There are no "real" InDesign scripting functions. Everything is bound to objects.
True. But from the perspective of a beginner, when you're starting, you have no idea what a function or a method is and you are just looking for "commands" to make stuff work. When I started with basil.js a few year ago and I had no idea about programming, I found it helpful after a while that I could always identify the b. as something, which was basil.js related.
But sure, if you all decided against it, there is not much I can do, apart from stating my opinion.

For the roadmap in general and for the stuff that will still be added to it, I hope I can be of help by giving you the perspective of somebody who started learn scripting as a total beginner with basil.js, so I think I can give some good insights in the initial confusions that I faced, when I first used basil.js. :) Hope to find some time in the coming weeks to add some points to the road map.

What should the procedure be? Should I introduce and discuss new ideas in a new issue? And if so, one issue per idea? Or should I group ideas?

fabiantheblind:
I would suggest grouping them in one issue. Then one of the core members can decide to integrate them.

fabiantheblind:
@trych also maybe we should exchange e-mail for easier discussion fabiantheblind@gmail.com

b-g:
Hi guys! 

Great to see some action here!! 

Getting rid of the *b.* seems to me like a good idea to have maximum compatibility with p5.js ... however I don't see a reason why we shouldn't have both ways of writing basil scripts in peaceful co-existence. Basically the idea is to inject into the "local" scope of setup() and draw() all api functions, but in the global scope the user will still have to use b.something. 

Best, B

b-g:
Merged roadmap.txt into master, to facilitate discussion! But would recommend to open a dev branch for everything else ... and to merge only once the feature/milestone is shipped with a new release.

fabiantheblind:
> Merged roadmap.txt into master, to facilitate discussion! But would recommend to open a dev branch for everything else ... and to merge only once the feature/milestone is shipped with a new release.

Yes of course. Each feature should have is own branch (IMHO).

PR #60, opened 2016-07-18
b.units() improvements, MM as default

trych:
Ok, this is my first ever PR on Github, please let me know, if am doing something wrong. Still learning.

Don't get immediately scared by the huge number of additions and deletions, they are caused by the yuidoc generation and I suspect (well, I hope) they happened, because yuidoc was updated since it was last used here.

Let the discussions begin.


includes/environment.js
```diff
@@ -83,7 +83,7 @@ pub.pasteboard = function ( h, v ) {
 };
 
 /**
- * Returns the current page and sets it if argument page is given. Numbering starts with 1. 
+ * Returns the current page and sets it if argument page is given. Numbering starts with 1.
```
trych: Ok, I don't get it, apparently.
Why does the PR split view suggest for many lines (like this one), that it was updated although I did not touch it and it has seemingly the same content. What am I doing wrong?

(See screenshot)
splitview

fabiantheblind:
Where did you actually do a change? You don't need to commit everything. You can also commit only one file. 

    git commit mychangedfile.js -m "changed some stuff"

fabiantheblind:
Please close this PR and just commit the change for the units. Did you run the tests?

trych:
Yes, this turned out a bit messy. Thanks for opening the yuidoc issue. I will recreate the PR in a clean way and leave out the yuidoc generation for now.

PR #62, opened 2016-07-19
Unit improvements, changed default unit to MM

trych:
This changes the default unit to MM and adds other minor improvements.

NOTE: I did run the test suite and it does _fail_ two of the tests in `/test/transform-tests.jsx`: `testHeight` and `testSize`.
However, after further testing, I think this might be due to some rounding error. When I insert some writeln statements before the failing assertions, it shows me that in fact it does, what it is expected to do. Try for example `/test/transform-tests.jsx` with these inserted writeln statements in the `testHeight` function:

```JavaScript
  testHeight: function(b) {
    var doc = b.doc();
    // create box at 0,0
    var rect = b.rect(0,0,100,100);

    $.writeln(b.itemHeight(rect)); // --> 100
    $.writeln(b.itemHeight(rect) === 100); // --> false - *should* be true, I guess?

    assert( b.itemHeight(rect) === 100 ); // --> fails here
    // move to 50,0
    b.itemHeight(rect, 50);
    // check new x coordinate
    $.writeln("BBB b.itemHeight(rect): " + b.itemHeight(rect));
    assert( b.itemHeight(rect) === 50 );
  },
```

This looks like a rounding error to me. In `testWidth` for example this does not happen. Maybe InDesign is calculating internally with PT and then has rounding errors happening sometimes when working with MM. Fact is, when you exit the script right before the failing assertion a rect was created which is exactly 100 mm x 100 mm large, according to the UI.

### DIFF
--- a/includes/core.js
+++ b/includes/core.js
@@ -215,7 +215,7 @@ var setCurrDoc = function(doc) {
   currLeading = currDoc.textDefaults.leading;
   currKerning = 0;
   currTracking = currDoc.textDefaults.tracking;
-  pub.units(pub.PT);
+  pub.units(pub.MM);
   
   updatePublicPageSizeVars();
 };
--- a/includes/environment.js
+++ b/includes/environment.js
@@ -499,11 +499,11 @@ pub.nameOnPage = function(name) {
 };
 
 /**
- * Sets the units of the document (like right clicking the rulers). The default units of basil.js are PT.
+ * Sets the units of the document (like right clicking the rulers). The default unit setting of basil.js is MM.
  *
  * @cat Document
  * @method units
- * @param  {Constant} [units] Supported units: PT, PX, CM, MM or IN
+ * @param  {Constant} [units] Supported units: MM, PT, PX, CM, IN
  * @return {Constant} Current unit setting
  */
 var unitsCalledCounter = 0;
@@ -538,7 +538,7 @@ pub.units = function (units) {
     currUnits = units;
     updatePublicPageSizeVars();
   } else {
-    error("b.unit(), not supported unit");
+    error("b.unit(), invalid unit. Please use: P.MM, P.PT, P.PX, P.CM or P.IN.");
   }
   if (unitsCalledCounter === 1) {
     warning("Please note that b.units() will reset the current transformation matrix."); 
--- a/includes/environment.js
+++ b/includes/environment.js
@@ -536,7 +536,7 @@ pub.units = function (units) {
       verticalMeasurementUnits = unitType;
     }
     currUnits = units;
-    updatePublicPageSizeVars();
+    if (unitsCalledCounter) updatePublicPageSizeVars();
   } else {
     error("b.unit(), invalid unit. Please use: P.MM, P.PT, P.PX, P.CM or P.IN.");
   }
--- a/test/transform-tests.jsx
+++ b/test/transform-tests.jsx
@@ -68,35 +68,43 @@ b.test('TransformTests', {
 
   testWidth: function(b) {
     var doc = b.doc();
+    // use tolerance value to avoid rounding errors
+    var tolerance = 0.0001;
     // create box at 0,0
     var rect = b.rect(0,0,100,100);
-    assert( b.itemWidth(rect) === 100 );    
+    assert( b.itemWidth(rect) > (100 - tolerance) && b.itemWidth(rect) < (100 + tolerance) );
     // move to 50,0
     b.itemWidth(rect, 50);
     // check new x coordinate
-    assert( b.itemWidth(rect) === 50 );    
+    assert( b.itemWidth(rect) > (50 - tolerance) && b.itemWidth(rect) < (50 + tolerance) );
   },
 
   testHeight: function(b) {
     var doc = b.doc();
+    // use tolerance value to avoid rounding errors
+    var tolerance = 0.0001;
     // create box at 0,0
     var rect = b.rect(0,0,100,100);
-    assert( b.itemHeight(rect) === 100 );    
+    assert( b.itemHeight(rect) > (100 - tolerance) && b.itemHeight(rect) < (100 + tolerance) );
     // move to 50,0
     b.itemHeight(rect, 50);
     // check new x coordinate
-    assert( b.itemHeight(rect) === 50 );    
+    assert( b.itemHeight(rect) > (50 - tolerance) && b.itemHeight(rect) < (50 + tolerance) );
   },
 
   testSize: function(b) {
     var doc = b.doc();
+    // use tolerance value to avoid rounding errors
+    var tolerance = 0.0001;
     // create box at 0,0
     var rect = b.rect(0,0,100,100);
-    assert( b.itemSize(rect).width === 100 && b.itemSize(rect).height === 100 );    
+    assert( b.itemSize(rect).width > (100 - tolerance) && b.itemSize(rect).width < (100 + tolerance) );
+    assert( b.itemSize(rect).height > (100 - tolerance) && b.itemSize(rect).height < (100 + tolerance) );
     // move to 50,0
     b.itemSize(rect, 50, 50);
     // check new x coordinate
-    assert( b.itemSize(rect).width === 50 && b.itemSize(rect).height === 50 );    
+    assert( b.itemSize(rect).width > (50 - tolerance) && b.itemSize(rect).width < (50 + tolerance) );
+    assert( b.itemSize(rect).height > (50 - tolerance) && b.itemSize(rect).height < (50 + tolerance) );
   }
 
   // todo: add matrix transformation tests here ...
@@ -105,5 +113,4 @@ b.test('TransformTests', {
 
 // print collected test results
 b.test.result();
-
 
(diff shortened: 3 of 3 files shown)


fabiantheblind:
Did you check what happens if the rect has no stroke only a fill?

trych:
Yes, I did (was also my first suspicion): Same thing happens unfortunately. I mean it does _not_ fail the `testWidth` test, so I think it must somehow be a rounding error, because it's doing pretty much the same stuff.

fabiantheblind:
Can you sort out this error before we merge this?

trych:
I'm afraid if this is an InDesign rounding error, there is not much we can do, unless we modify the testing script.
If I add some (ridiculously small) tolerance value to see, if the result is within that tolerance, it does not fail:

```JavaScript
  testHeight: function(b) {
    var t = 0.0000000000001;
    var doc = b.doc();
    // create box at 0,0
    var rect = b.rect(0,0,100,100);
    assert( (b.itemHeight(rect) > (100 - t)) && (b.itemHeight(rect) < (100 + t)) );    
    // move to 50,0
    b.itemHeight(rect, 50);
    // check new x coordinate
    assert( b.itemHeight(rect) === 50 );    
  },
```
It does not fail with:
`var t = 0.0000000000001;`
It does fail with:
`var t = 0.00000000000001;`

Looks like a rounding error to me. And again, when I interrupt the script after creating the rectangle, it has the expected size exactly from a UI perspective. And also the logic of the `$.writeln()` statements above indicates that it must be something like that:

```JavaScript
$.writeln(b.itemHeight(rect)); // --> 100
$.writeln(b.itemHeight(rect) === 100); // --> false (???)
```

So generally, it creates the shapes as expected, but fails the assertions due to a rounding error.

We now have three options:
1.) Modify the testing scripts.
2.) Leave PT the default value.
3.) Ignore the errors.

I don't like option 3 and I also think MM would be a better default (or who can tell me the point size of an A4 sheet? ;) ), so what do you think about option 1?

fabiantheblind:
I'm also for to option 1. Would love to hear from the other members about that. @ludzeller @ffd8 can we merge this? (Benedikt is on vacation right now.)

@trych can you:
- modify the test
- rename the `t` to `tolerance`
- add a comment about the rounding error
- add another commit for it to this PR

trych:
Btw., could somebody else check, if this fails in other versions, too? I am using CS6 under OS X 10.8.5. You could just quickly change the `PT` to `MM` in [includes/core.js, line 218](https://github.com/trych/basil.js/blob/lerpShape/includes/core.js#L218) and then run `/test/transform-tests.jsx`.

For my system this results in
```
6 test cases run
10 assertions run
8 passed
2 failed
```
Do you get the same results on other systems/with other ID versions?

fabiantheblind:
I'm currently also on CS6 only. My CC subscription went out and I'm not sure if I want to prolong it.

trych:
Ok, what the heck? I tried add another commit to the pull request, but for some reason git stated `1 file changed, 1 insertion(+), 1 deletion(-)`, although I changed several lines. I went ahead and on Github in the diff view it also only showed 1 line difference.

However, the resulting file in my branch looked fine, just like my changed file.

Upon further investigating I come to the conclusion that git seems to think that the entire file is one line? (and therefore Github shows me only the diff of one line, although the file looks fine).

Is this maybe connected to the "[Utf-8 Bom Commits Issues](https://github.com/basiljs/basil.js/wiki#utf-8-bom-commits-issues)" mentioned in the Wiki and the fact that this is a `.jsx` file? I don't know what a Bom is. Does anyone know, how to fix this? Deleted the commit for now.

fabiantheblind:
Have you tried to apply the fix mentioned in the wiki?  

    # open file in vim editor in the commandline 
    vim file.jsx
    # within vim do a : this is the command marker
    :set nobomb
    # wq writes out the file and quits vim
    :wq

fabiantheblind:
Ran tests in CC 2014& CC 2015 Osx 10.11.5:

    6 test cases run
    10 assertions run
    8 passed
    2 failed

trych:
Alright, thanks, so same rounding issues in other versions apparently.
I did try the vim thing, but I realize my mistake now, I accidentally copied `vim file.jsx :set nobomb :wq` as one line into bash instead of using the last two commands in vim. And then it naturally gave me some weird error in vim. Will try again later and then have one dedicated commit to get this file to work again and one to make the proposed changes.

trych:
I tried the vim thing now again, it still does not work, see [most recent commit](https://github.com/basiljs/basil.js/pull/62/commits/62c3089e42a81abc631fabdcad57aa222ad979af).

Here are my shell message afterwards:

```bash
trychs-mac:bundle trych$ git diff
warning: CRLF will be replaced by LF in test/transform-tests.jsx.
The file will have its original line endings in your working directory.
diff --git a/test/transform-tests.jsx b/test/transform-tests.jsx
index d6913d6..c2c09bc 100644
--- a/test/transform-tests.jsx
+++ b/test/transform-tests.jsx
@@ -1 +1 @@
    assert( b.itemSize(rect).width === 100 && b.itemSize(rect).height === 100 );    assert( b.itemSize(rect).width === 50 && b.itemSize(rect).height === 50 );  b.test.result();ed test resultsation tests here ...
\ No newline at end of file
    assert( b.itemSize(rect).width === 100 && b.itemSize(rect).height === 100 );    assert( b.itemSize(rect).width === 50 && b.itemSize(rect).height === 50 );  b.test.result();ed test resultsation tests here ...
```

And then after staging (and again the same after committing:
```shell
trychs-mac:bundle trych$ git add -A
warning: CRLF will be replaced by LF in test/transform-tests.jsx.
The file will have its original line endings in your working directory.
```

trych:
Aha! Finally I found something on the issue. See [this StackOverflow thread](https://stackoverflow.com/questions/1889559/git-diff-to-ignore-m). Now I just need to make sense of it.

trych:
Ok, now I got it to work! I replaced the `^M` line breaks with regular line breaks in `vim` with a command I found [here](http://stackoverflow.com/a/811212/3399765).

```vim
:%s/<Ctrl-V><Ctrl-M>/\r/g
```
Always learning something new!
So in b6932f9 only the line breaks get fixed to re-enable the file for proper git and github diff treatment and in 2283707 I do the actual changes on the test file to introduce the tolerance value.

We have to be somewhat careful in the future that the problematic `^M` are not re-introduced again.

This is ready for review at last. :)

b-g:
Hi @ffd8, are you fine with default units set to mm? Please confirm before I merge. Thanks! B

b-g:
Hi all! 

Had some time in the last couple of days to think over the unit issue (long walks in the forest :). I have to say that I changed my mind. I no longer think it is a good thing to change the default units to MM. 

IMO basil is a port of processing API wise and also thinking wise. Hence I believe we should aim for the following priorities (in that order):

* basil is for edu and teaching
* basil code examples, the basil API should be as close as possible compatible with processing (and p5js)
* basil is an computational design toolkit for making books / graphic design

Changing the default units would make the transfer of someone from processing to basil harder ... And in addition we would have to adjust potentially every existing code example / tutorial. Hence I'm voting against it.

(But I would like to keep of course the new tests and the better warnings of this PR)

@ffd8 what do you think?

fabiantheblind:
hm.  

> basil is an computational design toolkit for making books (…)

- Books are made on paper and for that we measure in MM.  
Question: What is the primary output medium for your students when using Basil?

trych:
Hm. My 2 cents: I actually have to agree with Fabian here.
~~If I want to stay purely digital, I could just use Processing. The moment I choose basil.js over Processing is when I want to code something that I print later.~~ (Well, come to think of it, you would also want to use it, when you want to stay digital, but need sophisticated typography functionality). Checking the official basil.js gallery almost all projects presented there are print project, so my impression that it is mostly used in print seems to hold true somewhat. IMHO InDesign is actually really awkward to use with PX, even if you design for web (no pixel snapping for example).

I can totally agree to stay as close to Processing as possible, but I would add "… where it makes sense". I think we should actually somewhat acknowledge the fact that we are in InDesign and that it is mainly used for print.

But as I said, that's just my 2 cents, after all, it would still be easy enough to switch to MM in your scripts.

ffd8:
Fun topic for which I too have formed  the following opinions:

Let's set b.units or pub.units from the start as a getter for whatever the current doc units are! 

In previous classes I've had indesign pros go crazy over the default changes (pasteboard being an issue too)- then I finally ran into my own issues while preparing a project. 

Screen != always PX && print != always MM (ie. USA -> IN)... 

I think as many of these settings as possible should be opt in. So units can be whatever the person set it to, we can suggest they use pixels if porting directly from processing.. But there is no preferred use for basiljs. Some will use it for books, others for onscreen 1000 page PDFs... Let's try to keep it as open to the users workflow.

In worst case, we could also create a sort of b.safeMode() or similar wording for total beginners with a few useful presets to reduce rendering errors.

trych:
This sounds pretty good to me.
Didn't even think about the MM / IN issue, euro-centric ignorant that I am. ;) This just means we would need to update many of the example scripts so they behave in a predictable manner.

fabiantheblind:
> Let's set b.units or pub.units from the start as a getter for whatever the current doc units are!
totally agree with that. Catching the default settings and restoring them at the end (if we need to change something) should be the way.

b-g:
Sounds like a good compromise! 👍

trych:
Great, so I will just close this PR, as we don't really need it anymore. Will do a separate PR for the one improvement that was independent of the unit issue (and that shall also teach me to only put things that really relate to an issue into the same PR ;) ).

PR #75, opened 2016-07-23
b.color() - bug fix and minor improvements

### LINKED ISSUE #73: b.color("colorName") error when using resulting color in b.fill() twice.
When `b.fill(myColor)` is used for the _second_ time with a color retrieved by `var myColor = b.color("colorName")`, InDesign throws an error `Color.===() kann mit Instanzen dieser Klasse nicht arbeiten` (sorry, I can't switch my installation to English unfortunately; something along the lines "Color.===() can't work with instances of this class").

I was digging around a bit, the error actually happens in the `checkNull` check that is performed at the beginning of `pub.fill()`. I suspect this is related to the fact, that `b.color("colorName)` returns a swatch instead of a color (as all other `b.color()` variants do, the error does not appear there). Possible fix might be to always return an actual `color` object from `b.color()`.

Here is a script to test (red rect is not really necessary, but makes it look more like a real world script):

```JavaScript

function draw() {
  var black = b.color("Black");
  var red = b.color(255, 0, 0);

  b.fill(black);
  b.rect(0, 0, 100, 100);

  b.fill(red);
  b.rect(100, 100, 100, 100)

  b.fill(black); // interrupts with error for me
  b.rect(200, 200, 100, 100);
}

b.go();
```

InDesign CS6, OS X 10.8.5.

Please flag as bug.

trych:
This fixes #73 by searching in the doc's color collection instead of the swatches. Made the argument checks and the colorMode checks stricter and improved formatting.

This passes `/test/all-tests.jsx`.

Please review.

### DIFF
--- a/includes/color.js
+++ b/includes/color.js
@@ -160,11 +160,11 @@ pub.color = function() {
   if (arguments.length === 1) {
     // get color by name
     if (typeof a === 'string') {
-      newCol = findInCollectionByName(currentDoc().swatches, a);
+      newCol = findInCollectionByName(currentDoc().colors, a);
       if (newCol) {
         return newCol;
       } else {
-        error("b.color(), a swatch with the provided name doesn't exist.");
+        error("b.color(), a color with the provided name doesn't exist.");
       }
     } else if (typeof a === 'number') {
       // GREY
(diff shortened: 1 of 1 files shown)


b-g:
@trych Many thanks for this bugfix! 👊 🐞  Merged!

PR #76, opened 2016-07-24
Fixed b.layer() bug to not set given layer to active

### LINKED ISSUE #74: b.layer(layer|string) does not set active layer to given argument
As the title describes.
~~It works fine for `b.layer()` and `b.layer("layerName")`, but for `b.layer(layerObject)` it does only _return_ the correct layer, but does not set it to active.~~

The method only _returns_ the layer and does not _set_ it (to the active layer of the doc), except when the layer is newly created.

Sample script:

```JavaScript

function draw() {

  b.clear(b.doc());

  var initialLayer = b.layer();
  alert(initialLayer.name); // --> Ebene 1; active layer stays Ebene 1

  // create layer "foo"
  var fooLayer = b.layer("foo");
  alert(fooLayer.name); // --> foo; active layer becomes foo

  var barLayer = b.layer(initialLayer);
  alert(barLayer.name); // --> Ebene 1; Ebene 1 becomes NOT active, foo is still active
}


b.go();
```
Please flag as bug.

trych:
- sets active layer now, if existing layer or layer name is given as argument
- made the docs more clear about the existing options
- added error message for wrong arguments

This fixes #74.

Please review.

Edit: And it passes all tests.

### DIFF
--- a/includes/environment.js
+++ b/includes/environment.js
@@ -314,7 +314,7 @@ pub.addToStory = function(story, itemOrString, insertionPointorMode) {
 };
 
 /**
- * Returns the current layer and sets it if argument layer is given.
+ * Returns the current layer if no argument is given. Sets active layer if layer object or name of existing layer is given. Newly creates layer and sets it to active if new name is given.
  *
  * @cat Document
  * @subcat Page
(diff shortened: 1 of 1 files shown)


b-g:
Hi @trych, many thanks for this! Merged! 👊 🐞

PR #80, opened 2016-07-25
b.image() - bug fix and added error handling

### LINKED ISSUE #49: Image placing with only x coordinate does not work as expected
As mentioned in issue #26 placing an image file (or .ai) does not work when using just the `x` coordinate as parameter.  

    function draw() {
      //b.image("test.png",0); // does not work as expected 0 should be x of newly created container
      b.image("test.ai", 0, 0, 100, 100); // works as expected
      var rect = b.rect(0, 100, 100, 100); // create a rectangle to place the image in
      b.image("test.ai", rect); // works as expected
    }
    b.go();

Attached is a test setup for this.  
[basil-place-ai-file.zip](https://github.com/basiljs/basil.js/files/349593/basil-place-ai-file.zip)

trych:
This fixes and improves b.image().

- there was no error handling, when wrong parameters were passed. Now, meaningful error messages are given: a `imgErrorMsg` explains possible options to the user. 
- there was a bug (that was not "issued" yet) that gave a wrong image size when using `b.image(img, x, y)` with `b.imageMode(b.CORNERS)`. This was fixed now, it now imports the image in its original size at the provided corner, as the method description would let you expect it.
- throws an error now if `w` or `h` are not greater than `0` when using `b.image(img, x, y, w, h)` with `b.imageMode(b.CORNER)`.
- updates the method doc to make more clear that you cannot use `b.image(img, x)`, but have to use `b.image(img, x, y)` (as there was some confusion about this). However, even if the user uses `b.image(img, x)`, then he gets the new meaningful error message.

Fixes #49.

This passes all tests. Please review.

### DIFF
--- a/includes/image.js
+++ b/includes/image.js
@@ -2,11 +2,11 @@
 // Image
 
 /**
- * Adds an image to the document. If the image argument is given as a string the image file  must be in the document's
+ * Adds an image to the document. If the image argument is given as a string the image file must be in the document's
  * data directory which is in the same directory where the document is saved in. The image argument can also be a File
  * instance which can be placed even before the document was saved.
  * The second argument can either be the x position of the frame to create or an instance of a rectangle,
- * oval or polygon to place the image in.
+ * oval or polygon to place the image in. If an x position is given, a y position must be given, too.
  * If x and y positions are given and width and height are not given, the frame's size gets set to the original image size.
  *
  * @cat Document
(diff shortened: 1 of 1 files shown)


fabiantheblind:
Awesome

b-g:
Hi @trych, many thanks for this! Merged! 👊 🐞

PR #81, opened 2016-07-26
Closes #76

GitBruno:
Check if swatch is indeed a colour.

### DIFF
--- a/includes/color.js
+++ b/includes/color.js
@@ -162,7 +162,11 @@ pub.color = function() {
     if (typeof a === 'string') {
       newCol = findInCollectionByName(currentDoc().swatches, a);
       if (newCol) {
-        return newCol;
+        if( newCol.hasOwnProperty("colorValue") ) {
+          return newCol; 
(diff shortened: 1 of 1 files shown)


trych:
Hi GitBruno,
I don't quite understand your commit. The swatch issue that fixes the "real world script" was addressed and fixed in #76 already. I changed it to only search the documents `colors`, which is what it is supposed to do, according to the description of the method. Can you explain why your commit changes the behavior to work with swatches again? Is there any advantage?

Thanks! :)

GitBruno:
Hi @trych I think you still need to merge the branch then, I just discovered that there was another instance of searching the swatches occurs. I like your solution so I have implemented it there as well. Cheers!

b-g:
Hi @GitBruno, many thanks for the PR! But I merged #75 by @trych which makes your one PR redundant! Any cheers for looking into it! Best, Benedikt

GitBruno:
Perfect!

Issue #42, opened 2016-07-06
Add size() function for setting canvas size


fabianmoronzirfas:
like Processing does it.

PR #83, opened 2016-08-22
feat(size): Added document size function

fabiantheblind:
Allows to get the current size as an Object set the width and height with one value or set the width
and height with two values

see issue #42 

test script:  

    function setup () {
      doc = b.doc();
      b.println('Get the size()');
      b.println(b.size().toSource()); // get the height and width
      b.println('set the size(100)');
      b.size(100); // set width and height to 100
      b.println('b.height: ' + b.height);
      b.println('b.width: ' + b.width);
      b.println('doc.documentPreferences.pageHeight: ' + doc.documentPreferences.pageHeight);
      b.println('doc.documentPreferences.pageWidth: ' + doc.documentPreferences.pageWidth);
      b.size(200,400);
      b.println('b.height: ' + b.height);
      b.println('b.width: ' + b.width);
      b.println('doc.documentPreferences.pageHeight: ' + doc.documentPreferences.pageHeight);
      b.println('doc.documentPreferences.pageWidth: ' + doc.documentPreferences.pageWidth);
    }
    
    function draw() {
    
    }
    b.go();

### DIFF
--- a/includes/environment.js
+++ b/includes/environment.js
@@ -19,6 +19,46 @@ pub.doc = function(doc) {
 };
 
 /**
+ * @cat Document
+ * @method size
+ * @param  {Number} width The desired width of the current document
+ * @param  {Number} [height] optional the desired height of the current document. If not provided the widht will be used as the height.
+ * @return {Object} if no argument is given it returns an object containing the current widht and height of the document.
+ *
+ */
+pub.size = function(width, height){
+  if(app.documents.length === 0){
+    // there are no documents
+    warning('b.size(width, height)', 'You have no open document.');
+    return;
+  } else {
+    if (arguments.length === 0){
+      // no arguments given
+      // return the curent values
+      // warning('b.size(width, height)', 'no arguments given');
+      return {width: pub.width, height: pub.height};
+    }
+
+    if(arguments.length === 1){
+      // only one argument set the first to the secound
+      height = width;
+    }
+    var doc = app.documents[0];
+    // set the documents pageHeiht and pageWidth
+    doc.properties =  {
+      documentPreferences:{
+      pageHeight: height,
+      pageWidth: width
+      }
+    };
+    // set height and width
+    pub.height = height;
+    pub.width = width;
+  }
+
+}
+
+/**
  * Closes the current document.
  *
  * @cat Document
--- a/includes/environment.js
+++ b/includes/environment.js
@@ -19,11 +19,15 @@ pub.doc = function(doc) {
 };
 
 /**
+ * Sets the size of the current document, if arguments are given.
+ * If only one argument is given, both the width and the height are set to this value.
+ * If no argument is given, an object containing the current document's width and height is returned.
+ *
  * @cat Document
  * @method size
  * @param  {Number} width The desired width of the current document
- * @param  {Number} [height] optional the desired height of the current document. If not provided the widht will be used as the height.
- * @return {Object} if no argument is given it returns an object containing the current widht and height of the document.
+ * @param  {Number} [height] optional the desired height of the current document. If not provided the width will be used as the height.
+ * @return {Object} if no argument is given it returns an object containing the current width and height of the document.
  *
  */
 pub.size = function(width, height){
--- a/scripts/examples/environment/size.jsx
+++ b/scripts/examples/environment/size.jsx
@@ -1,0 +1,41 @@
+#includepath "~/Documents/;%USERPROFILE%Documents";
+#include "basiljs/bundle/basil.js";
+
+
+// The b.size([width,[height]]) function allows to get/set the size of
+// the current document.
+//
+// If no argument is given it returns the current values of
+// b.width and b.height in an object {"width":Number,"height":Number}.
+//
+// If one argument is provided the width and height both get set this this value.
+//
+// If two arguments are provided the first value is the width the second is the height.
+//
+// Lets try this out:
+//
+function setup(){
+  // For starters - lets get the size
+  var documentSize = b.size();
+  b.println(documentSize.toSource()); // lets see what we've got
+  // to see the changes of that document we add a text in the exact size
+  b.text(documentSize.toSource(), 0, 0, documentSize.width, documentSize.height);
+  // next we pass only one argument
+  b.size(documentSize.width * 2);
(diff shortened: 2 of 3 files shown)


fabiantheblind:
tests included. All passing

b-g:
@fabiantheblind Yay! This is great! What about a simple example?

@ffd8 Ready to merge this (tomorrow if I don't hear a veto from your end)

fabiantheblind:
I'll write an example tomorrow.

ffd8:
Looks great! Sorry en route home the next 24 hr, so feel free to merge!

trych:
Looks good generally, two things I would add/change:

1.) To keep documentation of methods consistent, before the parameters there should be one short sentence added, describing what the method does. Like:

```Javascript
/**
 * Sets the size of the current document, if arguments are given. If only one argument is given, both the width and the height are set to this value. If no argument is given, an object containing the current document's width and height is returned.
 *
 * @cat Document
 * @method size
 * @param  {Number} width The desired width of the current document
 * @param  {Number} [height] optional the desired height of the current document. If not provided the widht will be used as the height.
 * @return {Object} if no argument is given it returns an object containing the current widht and height of the document.
 *
 */
```

You know, for beginners. ;)

2) Should the method not rather be called `b.docSize()`? I mean `b.size()` could refer to anything, `b.docSize()` would be much more descriptive and also go along nicely with the naming of `b.doc()`. What do you guys think?

Apart from that, it looks good. :)

ffd8:
I also wondered if b.size was too generic, but that's how processing deals with the canvas size so it makes sense.

Haven't been able to test it yet, but does it take into account the current b.UNITS setting, or just uses whatever the doc was set at?

fabiantheblind:
I also thought about how to name this function. Coming from the Processing side it is clear. `size()` is the same there. This makes code more portable. P5.js went the other way and named their "size" function `createCanvas`. In one way this is clearer. Saying exactly what it does. For me it always was a headache. While getting to know P5.js I always had to look up this function becahse the `size` call is already imprinted into my lizzard brain. I'm fine with both but would suggest keeping it more "processingish".  

@ffd8 it uses whatever is set by `b.units`

trych:
Ah, now I understand where the naming comes from. However, I wonder, if we should not depart from Processing's naming scheme, where it makes sense. Since there is no such thing as a document or a page in a Processing context, it makes sense that they did not name it `b.docSize()`. But in my opinion in an InDesign context such naming would make sense. It both describes what's going on and helps beginners as they can easily guess what's going to happen.

Having said all that, I think `b.pageSize()` would be even more accurate, as you actually set the dimensions of a page and with multi-page spread you have different total dimensions.

I guess we generally have to decide, if we want to strictly go with Processing's naming scheme or let the naming rather reflect InDesign's reality more. I would vote for the latter and therefore name this method `b.pageSize()`, but if others prefer to strictly follow the Processing nomenclature, then I could understand this, too.

fabiantheblind:
I could live with `docSize` even though  I prefer `size`. I would not use the `pageSize` proposal. It would lead to think you can change the size of one page. The `documentPreferences.{pageHeight:Number, pageWidth:Number}` gets applied to all pages in the document. Even after a page is created. To change the size of a single page you have to do something like in the code below. (These changes where intorduced to InDesign in CS5)  


```js
    var doc = app.documents.add(); // create a doc
    $.writeln(' bounds of  first page ' + doc.pages[0].bounds.toSource());
    var p2 = doc.pages.add(); // get a page before changes to docPrefs
    $.writeln(' bounds of  secound page ' + p2.bounds.toSource());
    doc.properties = {documentPreferences:{pageHeight:100, pageWidth:100}}; // affects the whole document
    // p2 now has the size of the doc not the preset one
    $.writeln(' bounds of  secound page after docPrefs change ' + p2.bounds.toSource());
    var p3 = doc.pages.add(); // add another
    var bounds = p3.bounds; // get the bounds
    // now resize that page only
    p3.reframe(CoordinateSpaces.INNER_COORDINATES, [[bounds[1], bounds[0]], [bounds[3] , bounds[2] / 2]]);
    // you can even transform the page 
    var matrix = app.transformationMatrices.add({counterclockwiseRotationAngle:45});  
    
    p3.transform(CoordinateSpaces.INNER_COORDINATES, AnchorPoint.CENTER_ANCHOR, matrix);  
    
    $.writeln(' bounds of p3 after changes ' + p3.bounds.toSource());
```

What would be cool would be a `pageSize()` function more or less like this.

```js

    function pageSize(page, [width, height]) {
      // TODO check if OD cs5 or higher
      // TODO check page type
      // TODO check second arg is array
      // TODO check if new size is smaller then the margins (creates an error)
      if (p instanceof Page) {
        if (arguments.length === 2) {
          if (arguments[1].constructor === Array) {
            var bounds = p.bounds;
            p.reframe(CoordinateSpaces.XXX, [[bounds[1], bounds[0]], [bounds[0] +
              width, bounds[1] + hright 2]]);
          }
        } else if (arguments.length === 1) {
          return p.bounds;
        }
    
      } else {
        warn('no page')
        return;
      }
    }
```

ffd8:
I agree that pageSize is confusing because changing size of one page is pretty rare in indesign... docSize makes more sense, but I still vote for just keeping it the Processing way. To be honest I think more people would set size from start or need an even fancier function for making a completely new doc and setting optional options (format, facing pages, etc)- but for just changing the size, I'd keep it the same. If Processing later changes to canvasSize than perhaps we should too.

Oh let's not forget fullScreen()! ... jk

trych:
True, `b.docSize()` is actually better than `b.pageSize()`. But sure, if you guys prefer `b.size()`, that's fine by me.

And about the `fullScreen()`: What do you guys think about a `b.screenMode()` to allow switching to presentation mode? This would be great for interactive scripts or installations.

fabiantheblind:
@ffd8 and @trych I actually like the idea of fullscreen. Should it be screenMode or fullscreen as in Processing? https://processing.org/reference/fullScreen_.html

b-g:
Hi all many thanks for the discussion! If I may summarize: `b.size()` it is. I'll merge it after launch (please veto is you really can't stand *size*).

b-g:
Thanks + merged!!

PR #87, opened 2016-08-24
Bundler

fabianmoronzirfas:
This PR has no changes to Basil itself. I just added my extendscript-bundlr so we can distribute Basil in a single file. 🎉  👯  😹 

We still need to decide how the bundled file should be named. Currently it is basil.bundle.js


trych:
As I am not familiar with "bundlers", could you explain to me, how this works? So there is some sort of automatism that grabs all the files in `includes`, merges them with `basil.js` into one large file `basil.bundle.js`? Is that right? How would I start this?

And if so, do you need to rebundle this then on every little change that we make? Or do we bundle it only on releases?

b-g:
Cool!

We should keep in mind not to break the include statement in all our files ...
```
```

Hence either `basil.js` becomes the bundled one, or if we minimise then `basil.min.js` (preferred) or if minimising is not possible then I'm happy with `basil.bundle.js`.

Did you include `basil.bundle.js`in `test/all-tests.bundle.jsx` https://github.com/basiljs/basil.js/pull/87/files#diff-efd1674055d345823d88bc270d4452dcR2 on purpose?

fabiantheblind:
@trych the tool is written in JS with Nodejs. It goes through the provided source file, resolves all paths and combines them into a new file (with lots of regex magic). To run this you need Node.js installed (v4.2.1 or higher). Then in the root of the basil git repo folder you run `npm install`and if everything went fine `npm run bundle`. Thats it.  

@b-g yes the includes in the examples are different referring to the version with all the include statements. I can take a look into minimising the bundle as well. I think we should keep the bundled version currently as an optional way to distribute Basil. Maybe for Basil 2.0 it could be the default. And yes. I added a test for the bundled version to see if all works as expected.

ffd8:
Chiming in to also re-itterate how important it is to keep the linking the #include statement the same with just "...basil.js" 

Regarding the whole bundler, I wonder if that should be held off for a 2.0 release? In which case, is there a benefit to keeping the files all separate as they are now in the includes? Yes it's organizationally nice.. but if the final goal is a single file– why not just make the switch to copy + paste it all into a single doc with a clear /* -- SECTIONS --*/ designation? Then there wouldn't be the need for anyone to keep up with bundling etc for each update. Would also help the hacking crowd who hates/disagrees with a function to quickly do a search and find the function for modding.. rather than looking through each file within the includes. I'm personally for the single basil.js file direction, but also dropping the need for an includes folder.

trych:
I'm not sure, if minimizing the bundle is a good idea.

When I was learning basil.js it was sometimes helpful to have a look at the source files to figure out, what is going on behind the scenes, without having to dig around on Github. Also, is there any advantage in minimizing it? I can understand that minimizing makes sense in a web context, but as basil.js is always just running locally, it should not really make any difference. Or am I missing something?

ffd8:
Agreed- that there's probably no benefit to minimizing it.. even if it was a few hundred kb download. Easily sifting through the code to learn from is more useful than shaving some space.

fabiantheblind:
I'm actually not sure if it is useful for development to have it all in one file. What if two people are working on different functions and create a pull request? This will end up in lots of conflict solving. Having it separated by topic is useful (IMHO).  
I would love to see the distribution version to be one single file but as @b-g pointed out. We will have to update all the examples and documentation before doing that. That's why I thought doing this for 2.0 would be good. As soon as there is only one file we can also tell the students that they can have their working directory whereever they want. Directly in the scripts panel or in their home dir. They only have to learn to create on relativ path to that.  



About minimizing. I'm with @trych Having the possibility to look into the source is awesome.

b-g:
Also some thoughts:

If it is working, I would also not hold it back and release it with the next basil.js release

What about having:
1. `./src` folder with nicely separated and readable files and `./src/basil.js` (basically everything what we currently have)
2. watcher process with gulp/grunt in background which bundles on changes in `./src` to
3. `./basil.js` = stripped out comments and whitespace ... as we should make indesign parse as little as possible

BTW. @fabiantheblind did you ever profile whether a single file vs. what we have now makes a difference?

fabiantheblind:
Just took a look around. There is a node module called watch [github.com/mikeal/watch](https://github.com/mikeal/watch). I could (for starters) integrate this into the build script. No need for other task runners.

fabianmoronzirfas:
What is the state of this PR?  
I took a look into the watch module I mentioned and it works great. So I would suggest making some of the changes as suggested by @b-g 

move everything into a `src` folder and compile it to ./basil.js

I would not remove all the comments. The comments help users to learn about basil. If you are okay with this I would make the changes and update this PR.  

So the structure would be:  


    .
    ├── basil.js (bundled version)
    ├── src
        ├──  basil.js
        └── includes
            ├── color.js
            ├── constants.js
            ├── core.js
            ├── data.js
            ├── environment.js
            ├── global-functions.js
            ├── image.js
            ├── math.js
            ├── private-vars.js
            ├── public-vars.js
            ├── shape.js
            ├── structure.js
            ├── transformation.js
            ├── typography.js
            └── ui.js

b-g:
Hi @fabianmoronzirfas, 

Thanks! Structure looks good to me! 

Comments:
1. is it faster?
2. what about the dependencies in `scr/basil.js` to `/lib` e.g. `b.download()` (shell execute hacks)? Are they still working for both basil.js files?
3. how do deal with the problem of the bundled basil.js for new commits? Would then every PR come with a new bundled basil.js file?

Best,
B

PS. Workshop in SF was good!

fabianmoronzirfas:
> is it faster?

I did a quick test. Actually there there seems to be no difference.  

    // @include "../../basil.js";
    /* global b */
    function setup () {
      var doc = b.doc();
      b.clear(doc);
      for(var i = 0; i < 500; i++) {
        b.ellipse(0, 0, 10, 10);
      }
    }
    function draw() {}
    b.go();



    | No  | state            | bundled             | included            |
    | :-- | :--              | :-                  | :--                 |
    | 1   | without open doc | [Finished in 12.4s] | [Finished in 12.8s] |
    | 2   |                  | [Finished in 17.3s] | [Finished in 17.2s] |
    | 3   |                  | [Finished in 17.6s] | [Finished in 17.6s] |
    | 4   |                  | [Finished in 16.9s] | [Finished in 17.1s] |

> what about the dependencies in scr/basil.js to /lib e.g. b.download() (shell execute hacks)? Are they still working for both basil.js files?  

Yes it is working. Works when you reference it to the `src` or to the bundled version. The path to `download.sh` is an absolut path to the users home directory (which actually should not be like this ;-)). For having Basil truly portable there needs to be a fallback like:  

    if(File('./lib/download.sh').exists !== true){
    // create file next to basil.js and execute
    }

Same thing goes for the `stop.jsx`.

> how do deal with the problem of the bundled basil.js for new commits? Would then every PR come with a new bundled basil.js file?

It should. We should add a development information in the README like

install Node.js (maybe using [nvm](https://github.com/creationix/nvm/blob/master/README.markdown) or [brew](http://brew.sh/) or [chocolatey.org](https://chocolatey.org/))  

run

    npm install
    npm run watch # bundles all files in src int root of project

fabianmoronzirfas:
So guys.
- I moved all the files to a src folder. 
- renamed src/basil.js to src/main.js (so we don't get confused which is which)
- renamed the basil.bundle.js version to the basil.js (so the bundled version is now the one that gets included)
- added some files to .gitignore

I think this is ready to merge. Please review.

fabianmoronzirfas:
Added some information to the wiki how to run the bundler and some useful git hints 

https://github.com/basiljs/basil.js/wiki/Home/_compare/a35246cfd689cab8e2f933d733eac6191fb45b56...925ac8f396ba8e88e6355993c3dbfcb366654a2e

b-g:
@fabianmoronzirfas very cool! Looks good! Two comments:

1. in which branch should this go now?
2. could you add every time the file is bundled a version and date string like the p5 files has? Seel first line of https://cdnjs.cloudflare.com/ajax/libs/p5.js/0.5.3/p5.js

fabianmoronzirfas:
> in which branch should this go now?

until we decide how to branch into master

> could you add every time the file is bundled a version and date string like the p5 files has?

done. Will be generated from package.json version and current date. Currently macOS only.

b-g:
Thanks!! Great, ready to merge (to master) from my end! Will do today in the afternoon/evening ...

@ffd8 please veto if you have concerns

b-g:
@fabianmoronzirfas many thanks! finally merged!

fabianmoronzirfas:
Awesome. I'm really glad to have this. Next thing is making it truly portable (download.sh and stop.jsx).

trych:
> Added some information to the wiki how to run the bundler

Ok, I just tried to follow the instructions there and got a lot of errors upon running `npm install`, after installing `node` first via homebrew. Anybody knows what's going on?

bildschirmfoto 2016-09-11 um 02 23 27

fabianmoronzirfas:
Hm. Wired. Which version of nodejs is installed? And which version of npm?

fabianmoronzirfas:
After some reading it seems like a problem with npm itself. Please post your npm and node version.

fabianmoronzirfas:
And is there a npm debug log in the folder? Please post this as well

trych:
> Please post your npm and node version.

node version is `v6.2.2`.
npm I can't even get the version of (it would be just `npm -v`, right?), it gives me the same error as shown above.

> And is there a npm debug log in the folder?

Sorry, in which folder would I find that?

fabianmoronzirfas:
The npm-debug.log will be generated in the root of your repository or wherever you executed the npm command. But getting this error from running `npm -v` points to the npm problem. 

Possible solutions for this problem:   

- http://stackoverflow.com/questions/12594541/npm-global-install-cannot-find-module

fabianmoronzirfas:
This guide also might help.  

- https://johnpapa.net/how-to-use-npm-global-without-sudo-on-osx/

trych:
> This guide also might help. 

This one did the trick, thanks!
However, I don't fully understand how to use the bundler yet (sorry, never bundled anything before). I run `npm install` only ever once right?

And then `npm run watch` does what exactly? It seems to go into some idle state and wait for something? What do I need to do?

trych:
Aaaaaaah, what sorcery is this? `basil.js` updates *while* I work on the `src` files? I had no idea, now this is really cool.  :sunglasses:

So, I take it that I have to run the `npm run watch` command before I start working on the `src` files and then when I am done, I can cancel it with `ctrl + C`, correct?

fabianmoronzirfas:
It's JS white magic.  :tophat: :rabbit2:
 
> So, I take it that I have to run the npm run watch command before I start working on the src files and then when I am done, I can cancel it with ctrl + C, correct?  

Yes that's right

trych:
And one more question. After running `nmp install` (I guess?), there is now this `node_modules` folder in my `bundle` folder, so naturally git wants to add that. However, I neither see this in master nor in `.gitignore`. Should this folder usually be somewhere else? Or why is not in master? Should we add it to `.gitignore`?

fabianmoronzirfas:
Oh. Yes. It needs to be in .gitignore.

PR #93, opened 2016-09-15
Make Basil.js fully movable

fabianmoronzirfas:
This PR makes Basil.js movable to other folders. The files

- lib/download.sh 
- lib/stop.jsx  

are created by basil when they are needed.  
What about the other files in lib/

- basil_simple.png
- basil.png

are they used somewhere in the library? Or are they just there for the docs or what not?  

This PR is only tested on macOS. A test for Windows would be great if the stop.jsx works as expected.


b-g:
Great!

Yes the images are used in the popup in the "hidden/offscreen" go modes.

fabianmoronzirfas:
Ah. Okay. So there is a decision needed for that. Currently if the script can't find the file it will ignore it and just create the progress panel. I can (and already have actually) implement a way to also create the image on the fly and use it in the panel, but that means adding a huge chunk of binary data into the source. This is a common technique for embedding image data into script ui. If you are okay with that I will commit. Take a look at [this gist](https://gist.github.com/fabianmoronzirfas/f814d2c90c23d7c0c36309228495dd8f#file-basil1-0-10-with-binary-image-js-L695) to see what it would be like. 

What is strange that the panel is sometimes not showing. Seems to be a problem with script ui. I remember that we had to add a `//@targetengine` to be sure it shows.  

I'm not so sure if we just should leave the image out like it is currently done.

trych:
Just my 2 cents:

I am not sure, if it really makes so much sense to stuff everything into one file, if it means that that file becomes cluttered and unreadable to some extent. Is there really any harm in having a small side folder with the file which hosts the external scripts and the images?

Also, we have to think this from the user side: What does the user actually download, where is he gonna move it, why would he move it somewhere else than suggested etc. There is also still the examples scripts folder. Is it not more messy, if the user starts moving around his `basil.js` file, while his sample scripts still sit somewhere else?

What actually goes in the user's download bundle at all at the current state? There have been a few changes recently, I think maybe we should take a moment and step back and do a clear and clean proposal about what's in that download bundle and how the user is supposed to use it, discuss pros and cons of the different options before we rush everything into one file.

fabianmoronzirfas:
The problem I see with the bundle folder is the distribution or lets call it selfcontainment of an project. We wrote these hacky installers to get rid of the process of "create a folder called basil, move the bundle there, create a folder called users, make a symlink into the scripts panel and so on"

First argument. In the current setting it is important to have the folder called bundle in the home directory. This is a hurdle I want to remove. With a fully bundled version I can tell my students:

- download just a single file from this URL
- place it in your project folder
- write `//@include "./basil.js"`

and they are good to go. Currently I never write the include statement by hand. I always have some boilerplate code around to copy paste for a new project.  

Second arg. If you want to "freeze" a project in its current state you just copy the basil.js file to your project folder and reference it relativ. 

Third arg. You can make a copy of basil.js and start to hack on the core as you like. No need for copying a whole folder.  

Forth arg. We could say for users "just download one file". For hackers "get the source from GitHub". The examples scripts could stay as a separate download.  
Fifth arg. Move the examples and tutorials out of the main repo. This would make contribution and development easier.  

I could go on. :-)

fabianmoronzirfas:
BTW I'm OOO for the next week and will only occasionally be able to respond on this thread.

ffd8:
Main benefit of suggesting a fixed location for the file (ie. documents/basiljs/bundle) is that one can open a template + a brand new doc in indesign and start testing away (students + I do it all the time).. no need to save the InDesign file first if you're just sketching or trying things out. If the basil.js single file sits in the project folder.. then one also runs into issues when 'packaging' the InDesign files (to gather all assets etc).

I do find the ability to save a versioned copy of basil.js w/ a project as a big plus– but again a static place is especially useful during a few hour workshop w/ people who have never touched Programming. I can already imagine the repetition of 'ok save your new file somewhere (w/ a folder), then manually copy the basil.js file (the right one) to that folder... If basil.js really reduces to a single js file (no attachments) then we remove the need for 'bundle' and simply throw it in the ~/documents/basiljs/ dir. Of course a tutorial/guide could also suggest to people the idea of loading it locally.. but tutorials/examples should all run without having to save the indesign file.

As for the two popup image files in lib vs. long base64 code.. I'd vote for removing the images altogether. I haven't used the hidden mode for a long time, so I forget what the UI feels like, but I'm guessing the pop-up could still be identifiable via ascii/textual formatting?

fabianmoronzirfas:
The benefit of a bundled file is (IMHO) that we _can_move it. You don't have to :)

fabianmoronzirfas:
> As for the two popup image files in lib vs. long base64 code.. I'd vote for removing the images altogether. I haven't used the hidden mode for a long time, so I forget what the UI feels like, but I'm guessing the pop-up could still be identifiable via ascii/textual formatting?

I totally agree with this one

fabianmoronzirfas:
So the current consensus is:  

- [x] don't add binary data for images
- [x] remove image fully from UI
- [x] leave the movement of Basil to the user
- [ ] move Basil from ~/Documents/basiljs/bundle to ~/Documents/basiljs?

Right? Could someone please still test on his ENV if the creation of the lib/download.sh or stop.jsx works.

b-g:
Hi Fabian, thanks for the nice summary and the PR! Wow, nice include hacks! I'm happy with everything, also moving out of the bundle folder. 

Tested it on my ENV in that order:
1. `bundle/scripts/examples/output/download.jsx` -> 👍 lib/ folder and download.sh was created (I just had to remove b.projectPath() in the example, seems this function does no longer exist? Maybe some forgotten legacy? Fine for me to remove it)
2. `bundle/scripts/examples/loop/animated nonsense ball.jsx` -> 👎, the stop script was not created in the lib/ folder

fabianmoronzirfas:
@b-g Hm  

> 2. bundle/scripts/examples/loop/animated nonsense ball.jsx -> 👎, the stop script was not created in the lib/ folder

did you move both of them to a new location and remove the old include statements? It works for me.   

create-lib-stop-jsx

b-g:
@fabianmoronzirfas thanks for the gif tut! :) Yes you were right, I didn't change all the paths ... but this shows also in what kind of trouble we potentially can get. If we change the paths this should be done in one big go. Right?

Should we merge this into dev or wait? Slightly confused ...

fabianmoronzirfas:
I would merge it. Moving the lib is optional. If the user moves it he will have to know that the path needs to be adjusted. It is still a tool for teaching programming. So making sure your paths are right is part of it I would say. I think it also depends on how we communicate this. I will not start with a template file for example. I will tell my students to type the setup draw include stuff by hand. This includes knowing where the lib is.

fabianmoronzirfas:
Lots of confusion here in Githubylon. :-) having the bundled version has nothing to do with moving the basil.js file to `~/Documents/basiljs/`. This is another thread. I just want the *possibility* to move the `basil.js` file. I had students where the estk didn't work at all or was pretty unresponsive (sometimes it hangs itself for several seconds). I also suggest to my students to move scripts into the `scripts panel` folder so they can be executed from the GUI. 
That's why I introduced the bundle. I know it is not as described on the homepage but IMHO easier. Another thing is that I teach it side by side with P5js. So they need to keep track of their paths anyway. 😝

trych:
> I had students where the estk […] was pretty unresponsive (sometimes it hangs itself for several seconds).

Yes, I had that problem as well in my seminar, this is due to Mac OS X's new «app nap». [Here's](http://blog.gilbertconsulting.com/2016/04/a-fix-for-extendscript-toolkit-running.html) how to fix it. However I also had some students where ESTK did not work at all, so I will offer Sublime as an alternative now.

fabianmoronzirfas:
> Yes, I had that problem as well in my seminar, this is due to Mac OS X's new «app nap». Here's how to fix it. 
Cool. Thanks for the hint. It is still not a good solution to disable app nap for the whole system because of a buggy and old ESTK. I also suggest Atom or Sublime Text as editor.

b-g:
@fabianmoronzirfas Merged :) + Many thanks for sorting it out!!

PR #95, opened 2016-09-18
Replacing findInCollectionByName()

trych:
This replaces the `findInCollectionByName()` helper function by generic (and faster) `itemByName()` methods, where appropriate. Since this does not work for styles (paragraph, character, object) as they can be nested, the old `findInCollectionByName()` was renamed to be a dedicated `findInStylesByName()`.

This fixes all the `b.[paragraph|character|object]Style()` methods that have been broken and therefore fixes issues #66, #67 and #68. It passes all tests.


b-g:
Hi @trych, Many thanks for this! Works on my end! PR looks good 👍! But as this is really nitty gritty detailed stuff ... would be good to have @fabianmoronzirfas 's confirmation (Fabian please merge if you are happy with it too). Cheers! Benedikt

fabianmoronzirfas:
Back from Italy with a nice tan. ☀️ 

I did a quick test and it worked for me. 👍  As we are already touching this what about styles with the same name? Currently it returns the first found item. Is that the desired behaviour?

Some remarks about the PR.  

1. (project structure) The PR should be against the develop branch not against master.
2. (coding style # 1) I suggest using the more verbose but safer if `(style !== null)` not `if(!style)` or `if(style.usValid)` as used in e.g. [line 513 of shape.js](https://github.com/basiljs/basil.js/pull/95/files#diff-44861868a123a96727f9d69b4086a9f3R513) or [316 of environment.js](https://github.com/basiljs/basil.js/pull/95/files#diff-51beb687548860e874dc662767ce8e6cR316). I'm not sure if these are typesafe.  
3. (coding style # 2) I also suggest dropping statements like `if(true) doSomthing;` in favour of `if(true){doSomthing}` (with breaklines of course). This is also a bit more verbose but better to read (IMO). Also eslint will warn about this.

trych:
Hi Fabian, welcome back!

> what about styles with the same name?

I went for the same behavior as `b.nameOnPage()`, which states in its refrence «Returns the first item on the active page that is named by the given name». I think this behavior makes sense. People should realize it's a bad idea to have paragraph styles with the same name and that this might be the reason their script does not work.

> The PR should be against the develop branch not against master.

Right. Sorry, I did not realize we already made the switch. Can a PR be moved to be against a different branch? Or do I just redo it?

> I suggest using the more verbose but safer if (style !== null) not if(!style) or if(style.usValid)

Alright for the `if (style !== null)`, the `if (style.isValid)` is a different type of check however. Using `itemByName()` *always* returns an object (that is not null), so the way you have to check, if it actually exists in the document is by using `if (style.isValid)`. Correct me, if I'm wrong.

> I also suggest dropping statements like if(true) doSomthing; in favour of if(true){doSomthing} (with breaklines of course).

Fine with me.

> Also eslint will warn about this.

What is the status about the linting issue? We had one person contribute a doc correction in the meantime that was turned down, because it also fixed the white spaces. The person did not commit another «non-whitespace-fix-PR». I think it is pretty annoying and demotivating for everyone if PRs are turned down because of stuff like this. Would it be possible to put this issue high on the list that it can be tackled pretty soon?

fabianmoronzirfas:
1. About returned elements: Okidoki. People should be punished if using same names for styles. Maybe we should close InDesign 😆 . (jk)
2. About the PR against develop: You can edit the PR and set the develop branch as comparison
3. About `isValid` your are right it works. I always thought this is not typesafe or also would check if the object just exist. I still prefer the more verbose version. We should use the same coding style. The rest of the lib is using the `!thing` and not `thing === true` so 👍  for `!thing`.  
4. About `if(){ }` cool.
5. about eslint. Yeah the eslint thing will take some more time. I dodging this work a bit because it means a lot of changes. :-/ But you are right. Needs a higher priority. Until I suggest using the `git add -p` interactive adding of changes actually I recommend using it all the time. Gives you a nice overview what you added. Still. I will make a PR for whitespaces only in the next days.

basil.js
```diff
@@ -1,4 +1,4 @@
-/* Basil.js v1.0.10 2016.09.08-10:54:10 */
+/* Basil.js v1.0.10 2016.09.19-02:45:55 */
```
fabianmoronzirfas: Nothing to fancy. Nothing to review

basil.js
```diff
@@ -1104,19 +1104,12 @@ var updatePublicPageSizeVars = function () {
...
-
-/*  var found = collection.itemByName(name);
-  if (!found || !found.isValid) return null;
-  return found;*/
-  
-   var found = null;
-   for (var i = 0; i < collection.length; i++) {
-     if (collection[i].name === name) return collection[i];
+// internal helper to get a style by name, wether it is nested in a stlye group or not
+var findInStylesByName = function(allStylesCollection, name) {
```
fabianmoronzirfas: This is all good.

basil.js
```diff
@@ -1733,9 +1726,9 @@ pub.addToStory = function(story, itemOrString, insertionPointorMode) {
   // init
   var libFileName = "addToStoryLib.indl";
   var libFile = new File(Folder.temp+"/"+libFileName);
-  addToStoryCache = findInCollectionByName(app.libraries, libFileName);
+  addToStoryCache = app.libraries.itemByName(libFileName);
```
fabianmoronzirfas: Why didn't anyone notice this is broken?

basil.js
```diff
@@ -1733,9 +1726,9 @@ pub.addToStory = function(story, itemOrString, insertionPointorMode) {
...
   // if and a cache is existing from previous executions, remove it
-  if (addToStoryCache) {
+  if (addToStoryCache.isValid) {
```
fabianmoronzirfas: is `isValid` a ID function?

basil.js
```diff
@@ -3868,9 +3861,8 @@ pub.strokeWeight = function (weight) {
  * @return {ObjectStyle}  The object style instance.
  */
 pub.objectStyle = function(name) {
-  var style = null;
-  var style = findInCollectionByName(name);
-  if(typeof style === 'undefined'){
+  var style = findInStylesByName(currentDoc().allObjectStyles, name);
+  if(!style){
```
fabianmoronzirfas: I prefer it more verbose.

basil.js
```diff
@@ -4197,8 +4189,8 @@ pub.color = function() {
 
   // check whether color was already created and added to colors,
   // keeps the document clean ...
-  newCol = findInCollectionByName(currentDoc().colors, props.name);
-  if (newCol) {
+  newCol = currentDoc().colors.itemByName(props.name);
```
fabianmoronzirfas: Did you see any changes in speed?

fabianmoronzirfas:
Sorry for the SPAM but this is actually a nice feature to review changes line by line. We also should consider making the develop and master branches protected so that they only can be merged when the PR is approved.

trych:
Do you expect me to react to all those or are you just testing out the feature? :D

fabianmoronzirfas:
Just testing how it looks like. :-)

b-g:
OK. It seems there is plenty of feedback here :) I have nothing to add ... @trych please give us a ping once you are ready to merge to /dev branch. Cheers!

trych:
> About isValid your are right it works. I always thought this is not typesafe or also would check if the object just exist. I still prefer the more verbose version. We should use the same coding style. The rest of the lib is using the !thing and not thing === true so 👍 for !thing.

@fabianmoronzirfas Sorry, I don't quite understand what's the decision now. `if (style.isValid) {doSomething}` and `if (!thing)`? So basically leave it as it was?

fabianmoronzirfas:
Yes leave it as is.

b-g:
@trych OK + But we still need a PR for the dev branch

trych:
Yes, and I will replace `if (true) doSomething` by `if (true) {doSomething}`, right?

fabianmoronzirfas:
Yes. Would be great

trych:
Okay, tried to get this ready at last – in the process I merged develop into the branch – now when I preview the merge back into the basil.js:develop it wants to merge a lot of changes. Should those not already be in the develop branch? Why does it want to update so much?

@fabianmoronzirfas Do you have any idea, what's going on?

fabianmoronzirfas:
Hm. Wired. What was the series of commands?   

    git checkout develop
    git pull origin develop
    git checkout findInStyles
    git merge develop

Would have been my guess

trych:
Yes, that's pretty much what I did.
It seems the PR now includes also those changes that Ted added a few days ago to master (our names for example), so I think I must have merged the master at some point, because I was not aware we already had made the switch to develop.

But then there is also some fixing of whitespace in there which I think you did yesterday in develop? I don't understand why this is added to the PR, this should be already in develop?

Dammit. Now how do I sort out this mess? 😞

trych:
Aha! Fabian, the whitespace fixes come from the fact that you had not bundled basil.js after you fixed the whitespaces in `src` file. So this PR comes with a bundled `basil.js` file where the whitespaces are fixed. 😎

So, when reviewing in diff view you can actually ignore the big basil.js file and skip right ahead to the files at the bottom.

Also, the changes that Ted made to master should probably also be included in develop. I guess we should generally avoid to go forward with both branches simultaneously, so we only ever update Master on releases. Is this correct?

So this can actually be reviewed (and is easy to review if you ignore the huge basil.js file with the whitespaces).

fabianmoronzirfas:
Thanks for sorting this out.

src/includes/core.js
```diff
@@ -451,10 +451,12 @@ var updatePublicPageSizeVars = function () {
 
 // internal helper to get a style by name, wether it is nested in a stlye group or not
 var findInStylesByName = function(allStylesCollection, name) {
-   for (var i = 0; i < allStylesCollection.length; i++) {
-     if (allStylesCollection[i].name === name) return allStylesCollection[i];
-   };
-   return null;
+  for (var i = 0; i < allStylesCollection.length; i++) {
+    if (allStylesCollection[i].name === name) {
```
fabianmoronzirfas: Thank you. When the linting is done some more of these statements will be removed.

trych:
Now this has merge conflicts, because the linting thing was merged first...

fabianmoronzirfas:
@trych argh. Will you resolve it? Or should I

trych:
Ok, good to go again. Somebody *please* merge this now.

fabianmoronzirfas:
> Somebody please merge this now.   

meh. lets wait another 12 hours or so…

fabianmoronzirfas:
just kidding. 😆

trych:
What a ride. Thanks!

PR #97, opened 2016-09-19
Adding ability to load files from absolute paths

rastapasta:
Hey team! Thanks for your nice project!

My current project needed access to specific system files to pull their returns to process it further - copying them before each script run wouldn't make sense in this case.

So I propose the inclusion of the ability to load files from outside the project's data/ folder! 😄 

Take a look at the simple + small change which makes it possible.

Thanks and cheers! 🍺

### DIFF
--- a/src/includes/data.js
+++ b/src/includes/data.js
@@ -718,6 +718,8 @@ var initDataFile = function(file, mustExist) {
   var result = null;
   if (file instanceof File) {
     result = file;
+  } else if (file.charAt(0) == '/') {
+    result = new File(file);
   } else {
     var folder = new Folder(projectFolder().absoluteURI + '/data');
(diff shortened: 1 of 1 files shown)


b-g:
Hi @rastapasta! 

Thanks for the compliments and the PR!
However this is already possible out of the box:

```
  // load file content
  // if you prefer not to use the data folder mechanism 
  // then use this code snippet:
  var txtFile = new File("/Users/bene/Desktop/some_test_file.txt");
  var text = b.loadString(txtFile);
```
Just use an instance of the `File` object to specify the file for loading in `loadString()`.

Will close this PR as our solution is lightly cleaner (think of the windows users) :)

Best,
Benedikt

rastapasta:
Hey Benedikt! Thanks pointing out that possibility!

Keep on with the great work,
- Michael

PR #103, opened 2016-09-28
fix(BOM and CR): fixed some utf-8 bom files with windows CR

fabianmoronzirfas:
When committing these files to git the linefeeds are not recognized so the whole commit creates one single line for the file.

This is maintenance only and can be merged right away.


b-g:
BOM Bang Bing! Thanks! Merged!

fabianmoronzirfas:
Cool. Just one thing. I went yesterday through the repo and deleted some PR branches. We should delete them right away after merging so the repo doesn't get to messy.

b-g:
@fabianmoronzirfas Thanks for the hint! Will from now on delete the branch right away (via github.com interface)

Issue #34, opened 2016-05-11
Enabling code linting? 


fabianmoronzirfas:
When writing scripts I normally use [eslint](http://eslint.org/) with the [Sublime linter package](https://packagecontrol.io/packages/SublimeLinter-contrib-eslint) for it. I also use the [Sublime Text CodeFormatter](https://packagecontrol.io/packages/CodeFormatter). Together they make up a strict and continuous coding style. Now a problem for eslint is the ExtendScript `#include 'path/to/file'` statement. It breaks the parsing. Fortunately there is an undocumented feature in ExtendScript for a different include statement:   

```
//@include 'path/to/file';
```

Or

```
// @include 'path/to/file';
```

Or even

```
//@include 'path/to/file';
```

Also the target application can be set like this. 

```
//@target 'InDesign';
```

This would allow:  
- to run eslint against user scripts
- to run eslint against the Basil source
- to enforce a strict coding style by adding a .eslintrc to the repo which contains rules suited for Basil. 

The next step would be running eslint against the whole project(examples and core). **Be aware that this might change a lot of the current formatting.** So I would do that but only if you guys want this to happen. 

First question for doing this is:
- single or double quotes? (It's some kind of religion 😏) I use single quotes. Throughout Basil there are double quotes. Single quotes make embedding double quotes for messages much easier. 
## How to approach the refactoring?

I think the best way would be formatting it file by file and test if nothing breaks. 

I would create a PR for each file.

fabiantheblind:
I did run eslint against the whole basil folder. Here is the result as HTML: 
https://dl.dropboxusercontent.com/u/1529019/basiljs/basil.eslint.report.html  

This is only a report. No changes where made.

some of the errors can be removed just by disabling the accompanying rules. e.g. no-undef comes from the fact that Basil is split into several files. Also ExtendScript has its `$` and `File` or `Folder` which are unknown to eslint. Others can be fixed just by running a code formatter. The indent for example.  

Others like e.g. "curly" need some manual adjustments.

b-g:
Hi @fabiantheblind,

Many thanks for the effort!!

I would find it super cool to have clean linted basil core files! 👍! Please go ahead (or @ludzeller any concerns?)! 

RE: single vs. double quotes
For me single quotes ... back then we were all new to the js shinny world ... and just went with the java/processing style, which we were familiar.

RE: `#include 'path/to/file'`
[...]

Issue #33, opened 2016-05-11
Questions on Basil.js


fabiantheblind:
Here there guys. I've been working with Basil for the last month and I was thinking of incorporating it deeper into my next seminar. I have some questions though:  

- How do I disable the error handling? It is great for beginners to not have the std ID error messages but to pinpoint problems it makes it harder. 
-  How do you test new features before merging them into master? It would be great to have some guidance for developers. 
- Is there a roadmap for basil? Where is it heading?
- Are you still using it in your curriculum?

Last but not least. You did a awesome job with it. 👍 Great work.

ludzeller:
Hello Fabian,

Great to hear that you like basil.js so much and that you would like to teach it in you class. That's the Interface Design course at FH Potsdam, right?

Regarding the error handling... The others might know more, but I think it's not configurable. Or could you elaborate a bit more on what you dislike and how you imagine it to be?

We have a set of automatic tests that we definitely run very often. And we use our gut feeling. Again, maybe I didn't fully get your question... :) Besides the tests there is no strong protocol to follow, just be careful and always think in the perspective of programming beginners. We take them seriously.

Our roadmap is quite simple... we as the existing members are currently not planning to extend basil.js further. We have the feeling that we reached a solid spot where the most essential features are already present and that the InDesign (JavaScript) engine has limits that prevent e.g. a fast real-time mode. 

But we are still using it in our seminars and will immediately fix it, if Adobe updates break it. But  it's currently not as intense as it obviously had to be when we developed it. For three semesters around 2013 we had been very busy incorporating it in our seminars, also to find the time to build it and test it in real-life  teaching. Right now, it's more living as a side option, so we do not focus full seminars on it but rather teach it together with Processing (obviously) as a side option for generative print.

But we are always open for new ideas and contributors! And other lecturers who want to do classes about it. So, if you need any basil.js help for preparing your class, please let us know!

trych:
Hi Fabian,
great to see that there is some stuff going on with basil.js still. I was also teaching basil.js in a brief introduction seminar two semesters ago (at HAW Hamburg) and the idea is to continue with this next semester.

I recently also collected a few ideas for additions and improvements to basil.js and was looking, if I could implement them myself at some point, but as I just finished my Bachelor thesis (making heavy use of InDesign scripting of course ;) ), I did not find the time yet, to post my ideas here in an organized manner. I will hopefully get around to that in the next weeks.

Looking forward to bringing the project forward!
Timo

[...]

PR #104, opened 2016-09-29
starting linting

fabianmoronzirfas:
So i did some linting on the first file. `color.js`
Ran the tests and all are passing.  
I did not commit a build of the bundled version so we don't run into conflicts until PR #95 is merged.  

# Rules to consider

---  

[consistent-return - Rules - ESLint - Pluggable JavaScript linter](http://eslint.org/docs/rules/consistent-return)

States that functions should always return a value or never. Not some do some dont. Actually I think it would be good to go for a always return value. Could be:  

    return 0;

if nothing the users needs comes back. Currently functions without return statement return `undefined`.  

---  

[max-len - Rules - ESLint - Pluggable JavaScript linter](http://eslint.org/docs/rules/max-len)

Lines should not exeed a specific length.  

I'm not sure what happens with the comments for yuidoc. Anyone knows what happens with breaklines?  

if it doesn't create any problems I would say lets go for a max length of 80 or 100 characters. Lines stay more readable like this.  

---  

[no-lonely-if - Rules - ESLint - Pluggable JavaScript linter](http://eslint.org/docs/rules/no-lonely-if)  

> disallow if statements as the only statement in else blocks 

Example:  

Wrong

    if (condition) {
        // ...
    } else {
        if (anotherCondition) {
            // ...
        }
    }

Right:  

    if (condition) {
        // ...
    } else if (anotherCondition) {

    }

---  

[no-else-return - Rules - ESLint - Pluggable JavaScript linter](http://eslint.org/docs/rules/no-else-return)

Example 

[basil.js/color.js at master line 183](https://github.com/basiljs/basil.js/blob/master/src/includes/color.js#L183)  

Currently it is:  

    if (typeof a === "string") {
      newCol = findInCollectionByName(currentDoc().colors, a);
      if (newCol) {
        return newCol;
      } else {
        error("b.color(), a color with the provided name doesn't exist.");
      }

And it should be:  

    if (typeof a === "string") {
      newCol = findInCollectionByName(currentDoc().colors, a);
      if (newCol) {
        return newCol;
      } else if(newCol === null) {
        error("b.color(), a color with the provided name doesn't exist.");
      }

The some goes for [basil.js/color.js at master line 329](https://github.com/basiljs/basil.js/blob/master/src/includes/color.js#L329)  
The some goes for [basil.js/color.js at master line 513](https://github.com/basiljs/basil.js/blob/master/src/includes/color.js#L513)  

---  

I currently set these rules to warn instead of error. I will provide a write up on how to configure sublime and eslint to have a proper environment.


trych:
> Could be: `return 0;`

Is returning `0` not problematic in cases, where `0` can be a valid output (and for example `null` wouldn't be. I.e. would it not be better to always return `null` instead? Or am I misunderstanding something?

And I don't quite understand the last point. Does the error count as a return statement here?
And should it then not rather be

```JavaScript
if (typeof a === "string") {
  newCol = findInCollectionByName(currentDoc().colors, a);
  if (newCol) {
    return newCol;
  } 
  error("b.color(), a color with the provided name doesn't exist.");
```
after linting?

fabianmoronzirfas:
## About consistent return values  

In languages like c++ it is normal to have a return value of 0 when everything went fine. If there was an error you return 1 or a error number. After reading the eslint docs I'm a bit confused myself. I think I got it wrong.  

Having a

    return null;

at the end of the function solves the lining error. I think the rule is not about every function has to have a return statement but rather about if there is a return statement, then each "endpoint" should have a return statement.  

Actually it makes sense that have that consistent. It allows to always check if the returned value is what we need or null.

fabianmoronzirfas:
## About else return

In this case the snippet I provided is to short. We need to take a look at the whole function. The snippet just says

    if a is a string and I can find it in the collection return it if not warn the user. 
The else is right in that case but the `else if` is more accurate. The function still goes on if it is not a string.

fabianmoronzirfas:
We don't need to comply to all the rules eslint gives us. We can disable them. I actually like it to have the code a bit more verbose but easier to understand. These are a thing of style and also opinion.

trych:
Aah, now this has been merged, before I could give a comment on the cleaned code.

Few points that come to mind:

1. Did we not decide on single quotes instead of double quotes. I seem to remember that.
2. What about [line 314](https://github.com/basiljs/basil.js/pull/104/files#diff-2993dce13ed8da7b77ce266e06a3bf90R314)? I guess it is commented out, since `props` is not used anywhere. I would strongly vote for taking it out altogether then. Else there will be lot of zombie code left and later we are not sure if it needs to stay there or not.
3. About the last rule mentioned above: I still do not understand the two different versions that you are comparing. Could you provide the whole snippet after linting? I think if there are rules that we do not really understand or whose benefit we don't really understand, we should actually *not* turn them on.

fabianmoronzirfas:
@trych 
About quotes:  

I remember having already a discussion about that. @b-g and I where for single quotes @ludzeller made the point that Processing uses double quotes so porting code from Processing would be easier.  

here #34 and here #33  

About l314 props:

Yes you are right. I forgot about it. Sorry. I will include it in the next PR.  

About rules we don't understand and:  

Of course rules that we don't consider important for us should not be in there. I will take a deeper look into it and report back.  

@basiljs woohoo! linting is happening. I ❤️  it.

fabianmoronzirfas:
About this branch. I will use it to push more linting changes. So please don't delete it.

PR #114, opened 2016-10-31
Adding b.LOREM; placing it in two example scripts

### LINKED ISSUE #108: b.LOREM
Hi there,
after a little break here, I am back with lots of input, ideas and feedback from the course I am teaching. I will therefore open a few issues now, with suggestions for improvement. Btw, the course is going great, _lots_ of people learning programming from scratch and making nice progress.

The first one is really simple, but I think for a heavily typography related project like basil.js it would be nice to have a lorem ipsum constant, that simply contains the lorem ipsum string. In the course, demonstrating how to work with text I need to grab some text _all the time_ and I am getting tired of typing "Hello, I'm Timo." and "afölsdkjföalsdjfö aljdfölakjs". ;) Would be useful for sample scripts, too.

`pub.LOREM_IPSUM = "Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.";`

What do you think?

trych:
To get things going: the new `b.LOREM`.

I put it in two example scripts where I found it appropriate. There are 2 or 3 more example scripts with a lorem at the moment, but I think in those other cases we could use the upcoming `b.placeholder()` instead.

Closes #108.

### DIFF
--- a/basil.js
+++ b/basil.js
@@ -1,4 +1,4 @@
-/* Basil.js v1.0.10 2016.11.01-00:17:16 */
+/* Basil.js v1.0.10 2016.11.04-20:27:57 */
 /*
   ..-  --.- ..- -.... -..-- .-..-. -.-..---.-.-....--.-- -....-.... -..-- .-.-..-.-.... .- .--
 
@@ -33,6 +33,9 @@
   Please note: Big general parts e.g. random() of the basil.js source code are copied
   from processing.js by the Processing.js team. We would have had a hard time
   to figure all of that out on our own!
+
+  The Lorem ipsum string of b.LOREM is taken from https://indieweb.org/Lorem_ipsum and
+  is available under a CC0 public domain dedication.
 
   Supported Adobe InDesign versions: CS 5+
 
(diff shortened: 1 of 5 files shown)


b-g:
@trych Happy Halloween + Many thanks! Quick Q: where did you get the lorem text from? Don't want to run into any copyrights or licence issues ... Is it safe to use w/o credits etc?

trych:
Well, I found it in some old papyrus rolls by some roman dude I had lying around. 😉 

No, good point you raise, I found it on some Lorem Ipsum generator site, but apparently that site does not have any license attached to it, so you are right, I would *not* consider it safe to use. I did also not find any CC or Public domain instance of this specific lorem ipsum version anywhere else.

Therefore we will have to look for an alternative. What about the [Lorem ipsum version on the English Wikipedia page](https://en.wikipedia.org/wiki/Lorem_ipsum#Example_text)? Would that be safe to use?

In this case I would go for:

`Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.`

Is this okay?

trych:
Ok, down the rabbit hole of licenses. After giving the Wikipedia CC license a read, I think it is *not* compatible with basil's MIT-License (which btw. needs an update addressing the license year).

However, I have found the [same Lorem Ipsum text under a CC0 license](https://indieweb.org/Lorem_ipsum), which then would be totally fine to use.

b-g:
@trych Brilliant! Many thanks for going down the 🐰  ⃝  !! Plz add the wikipedia link or the indieweb reference in the js doc header, so that we have document it. Thanks! Will merge afterwards!

trych:
Ok, this is updated now. Sorry, this took a while. Is the license info in the right place?

b-g:
@trych Many thanks + looks good + merged! LOREM Amen 👍

PR #115, opened 2016-11-07
b.placeholder()

trych:
As discussed in #108 this adds `b.placeholder()` which can be used to fill textFrames with InDesign's semi-random placeholder text.

Usage is simply `b.placeholder(tf);`. It also returns the generated placeholder text, so you can do

```js
var tf = b.text("Initial text", 0, 0, 100, 100);
var ph = b.placeholder(tf);
b.typo(ph, "fillColor", b.color(255, 0, 0);
```

Added an example script as well, which creates these simple sample cases:

bildschirmfoto 2016-11-07 um 18 09 29

This passes all tests. Please review. 😎

### DIFF
--- a/basil.js
+++ b/basil.js
@@ -1,4 +1,4 @@
-/* Basil.js v1.0.10 2016.10.24-18:38:50 */
+/* Basil.js v1.0.10 2016.11.07-17:50:59 */
 /*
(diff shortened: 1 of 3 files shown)


b-g:
@trych Many thanks + very cool! Merged!

PR #124, opened 2016-11-08
Text collection fix

### LINKED ISSUE #116: b.words(doc), b.lines(doc) etc. are broken
(Last class there was a lot of stuff coming up that was broken/not working as expected, so I will have to add a few issues again)

`b.words()`, `b.lines()` etc. are not working when using `doc` as the container, although the documentation suggest they should. The (now broken) sample script [`forEach-loops.jsx`](https://github.com/basiljs/basil.js/blob/master/scripts/examples/structure/forEach-loops.jsx) suggests that it did work in the past.

The actual issue is [in this line](https://github.com/basiljs/basil.js/blob/master/src/includes/structure.js#L53), where it checks for the property `content`, which a document does not have. 

And while we're at it fixing this, I suggest we also allow `page` as a possible container.

trych:
To somewhat prove the point that I'm not only complaining and making up issues: Here's a bug fix. 😉

This fixes the bug with text collections when using a `document` as a container as described in #116. So stuff like `b.words(b.doc())` now works again.

Additionally I added `page` as a possible container.

Here is the changes I made:
- since all the textCollection functions were basically repetitive I created one generic helper function to handle all the functionality. The original textCollection functions now only pass the type of collection to this helper function. This is cleaner and will make future changes easier as we only have to change one function, not 5.
- in all the textCollection functions I renamed `item` to `container` for clarity.
- I replaced the function 'forEachStoryProperty()' with the simplified 'forEachTextCollection()' that also works with `page` as the container.
- I updated and improved the docs
- I created more meaningful error messages that state which containers are legal
- shortened the dummy text of the "forEach-loops.jsx" sample script (I used this script for testing and it runs ridiculously long with the original text, looping over all the characters).

I ran all tests, no fails. I also ran benchmarks, the new version is pretty much exactly as fast/slow as the old version.

This fixes #116.

Please review. 😎

### DIFF
--- a/basil.js
+++ b/basil.js
@@ -1,4 +1,4 @@
-/* Basil.js v1.0.10 2016.11.04-20:27:57 */
+/* Basil.js v1.0.10 2016.11.08-02:37:21 */
 /*
   ..-  --.- ..- -.... -..-- .-..-. -.-..---.-.-....--.-- -....-.... -..-- .-.-..-.-.... .- .--
 
@@ -1232,30 +1232,20 @@ pub.stories = function(doc, cb) {
 };
 
 /**
- * If no callback function is given it returns a Collection of items otherwise calls the given callback function with each paragraph of the given document, story or text frame.
+ * If no callback function is given it returns a Collection of paragraphs in the container otherwise calls the given callback function with each paragraph of the given document, page, story or textFrame.
  *
  * @cat Document
  * @subcat Multi-Getters
  * @method paragraphs
- * @param  {Document|Story|TextFrame} item The story or text frame instance to iterate the paragraphs in
+ * @param  {Document|Page|Story|TextFrame} container The document, story, page or textFrame instance to iterate the paragraphs in
  * @param  {Function} [cb]  Optional: The callback function to call with each paragraph. When this function returns false the loop stops. Passed arguments: para, loopCount
  * @return {Paragraphs[]} Array of Paragraphs.
  */
-pub.paragraphs = function(item, cb) {
+pub.paragraphs = function(container, cb) {
 
-  checkNull(item);
+  var legalContainers = "Document, Story, Page or TextFrame.";
+  textCollection("paragraphs", legalContainers, container, cb);
 
-  if(!item.hasOwnProperty("contents")) error("b.paragraphs(), Wrong object type.");
-
-  if(arguments.length === 1) {
-    return item.paragraphs;
-  } else if (cb instanceof Function) {
-    if (item instanceof Document) {
-      return forEachStoryProperty(item, "paragraphs", cb);
-    } else {
-      return forEach(item.paragraphs, cb);
-    }
-  }
 };
 
 // /**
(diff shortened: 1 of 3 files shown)


b-g:
@trych Many thanks for the bugfix and improvements! Before I merge ... can I get a green light from @fabianmoronzirfas for it too? I never really used the special extendscript "everyitem" stuff. Hence not the right person to review this.

fabianmoronzirfas:
@b-g @trych I'll review ASAP. I think tomorrow.

trych:
> could you add every time the file is bundled a version and date string like the p5 files has?

Hm, this somehow backfires now, as it will cause a merge conflict whenever another PR is submitted before previous PRs are merged, like in this PR here for example.

Any chance to take this out again? Or any way to work around this? Because it gets pretty annoying if for every PR you either have to wait for previous PRs to be merged to open it or alternatively solve a merge conflict.

I would like to at least take it out the develop branch. If we do a release of the master branch we could still put a date in.

fabianmoronzirfas:
Can you point to a conflict where this is happening? 

one way would be to

1. Don't commit the bundled file for a PR
2. Merge the changes of a PR
3. do a build on develop and commit there the bundle

b-g:
No strong feelings here. But I guess the easiest one would be just to have the timestamp in the official releases (in master branch). So I guess `npm run watch` would no longer add a timestamp ... but then we would need to introduce a `npm run release` to separate these things.

fabianmoronzirfas:
Looks good to me and works. I give it a 👍  and 💚  as well. @b-g Should I merge?

fabianmoronzirfas:
@trych Some remarks on the PR

1. I think it is easier if you create your branches in this repo and make a PR from your branch to develop
2. It would be good to separate the documentation changes from the code changes in the commit. This can be done by using the `git add -p` followed by a `git commit -m "my message"` command I mentioned in the [wiki](https://github.com/basiljs/basil.js/wiki#review-your-changes-before-committing).
3. Or you first do the code changes. If you are done do the commenting.  

This will make reviewing much easier.

b-g:
🎉 Happy to press the merge button (to feel more useful :)

fabianmoronzirfas:
@trych see also http://stackoverflow.com/questions/1085162/commit-only-part-of-a-file-in-git for more info

fabianmoronzirfas:
@b-g okay. 😄  next time I'll press that button or give you a call to do it 😉

trych:
Ok, will make the documentation and code changes separately in the future.

But why is it easier/better to create my branch in this repo here? I am not suggesting it isn't, but I just don't know git that well yet, so I am curious what the difference would be. 😊

fabianmoronzirfas:
> But why is it easier/better to create my branch in this repo here?

If you have a fork I need to clone the fork to review any changes. If it is here I just can fetch all new branches change to it. Maybe do a compare to different revisions. I think it is better to have the development under one roof. You are not a outside contributor

PR #132, opened 2016-11-13
Speeding up b.clear()

trych:
This PR speeds up `b.clear()` by avoiding the internal use of `b.forEach()`. A benchmark script `clearItems.jsx` is added to `test/benchmarks/`.

On my system the benchmark gave these results:

```
Old b.clear():
Clearing 500 grouped items: 8.066 seconds.
Clearing 500 items on page: 4.259 seconds.
Clearing 500 items on layer: 6.324 seconds.
Clearing 500 items in document: 2.749 seconds.
```

```
New b.clear():
Clearing 500 grouped items: 0.275 seconds.
Clearing 500 items on page: 0.589 seconds.
Clearing 500 items on layer: 0.561 seconds.
Clearing 500 items in document: 0.453 seconds.
```

The more objects need to be deleted, the greater the speed difference. It is also greater if the objects are deleted from the active page as the page needs to be redrawn more often with the old `b.clear()`.

It passes all tests.

### DIFF
--- a/basil.js
+++ b/basil.js
@@ -1,4 +1,4 @@
-/* Basil.js v1.0.10 2016.11.13-01:44:57 */
+
 /*
   ..-  --.- ..- -.... -..-- .-..-. -.-..---.-.-....--.-- -....-.... -..-- .-.-..-.-.... .- .--
 
@@ -1418,7 +1418,7 @@ pub.items = function(container, cb) {
 
 
 /**
- * Removes all PageItems in the given Document, Page, Layer or Group.
+ * Removes all PageItems (including locked ones) in the given Document, Page, Layer or Group. If the selected container is a Group, the Group itself will be removed as well.
  *
  * @cat Document
  * @method clear
--- a/basil.js
+++ b/basil.js
@@ -1,4 +1,4 @@
-/* Basil.js v1.0.10 2016.11.07-17:50:59 */
+/* Basil.js v1.0.10 2016.11.13-01:44:57 */
 /*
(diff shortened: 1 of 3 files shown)


fabianmoronzirfas:
BÄM 🚀

basil.js
```diff
@@ -1437,20 +1437,18 @@ pub.clear = function(container) {
...
-        // otherwise deletion process is confused
-      if(item !== null && !(item instanceof Group)) {
-        if(item.locked) error("b.clear(), some items are locked. Please unlock them first and sue then b.clear().");
-        item.remove();
-      }
-    });
+    container.pageItems.everyItem().locked = false;
+    container.pageItems.everyItem().remove();
+
+  } else if (container instanceof Group) {
```
fabianmoronzirfas: @trych Why are groups treated differently?

trych: Because in InDesign you cannot have a group without any pageItems in it. That means the group will not survive the clearing anyway, so you can delete the group right away, which also turns out to be faster according to the benchmarks.

fabianmoronzirfas: Okay. Does it also work if on the page are groups?

fabianmoronzirfas: Mybe I was not clear enough. if we use `b.clear(page)` and there are groups on that page with other objects. Will we run into a problem?

trych: No, not at all. As the container is going to be the page in this case, the groups that are contained on that page are just regular pageItems and will be simply removed like any other pageItems without any issues.

b-g:
@trych Many thanks + merged!

PR #137, opened 2016-11-20
Changed github names to real ones; logged b.clear() improvements

trych:
The switch from github names to real ones was only in master, therefore I recreated it in develop, so the branches are not out of sync.

Additionally I just started logging the changed features (well one, so far).

I think in the future we should make this a habit to always add the change to the changelog right away if we have a PR with a new feature. That way we don't have to collect every change before a release in a hurry next time. 😉


b-g:
@trych Thanks + Good call! Let's log the improvements/changes/fixes straight away from now on!

PR #141, opened 2016-11-25
Expanding functionality of styles

trych:
This greatly improves working with styles in basil.js.

1) All existing style functions (`b.paragraphStyle()`, `b.characterStyle()`, `b.objectStyle()`) are updated:

  * They can now either get a style by name or the applied style of a given textObject/pageItem. If a style of a give name does not exist, it gets created. Additionally a properties name/value pairs object can be used to set the new style's properties or to update an existing style's properties.

2) The new functions `b.applyParagraphStyle()`, `b.applyCharacterStyle()` and `b.applyObjectStyle()` are added. They take a textObject|textFrame|Story/pageItem and apply the given style to it. The style can be given as style instance or by name.

3) The example script `styles.jsx` is added to demonstrate how to use the style functions.

This passes all tests, please review. 🔍

### DIFF
--- a/basil.js
+++ b/basil.js
@@ -4776,12 +4776,35 @@ pub.textTracking = function(tracking) {
  * @param  {String} name      The name of the character style to return.
  * @return {CharachterStyle}  The character style instance.
  */
-pub.characterStyle = function(name) {
+pub.characterStyle = function(textOrName, props) {
+  var styleErrorMsg = "b.characterStyles(), wrong parameters. Use: textObject|name and props. Props is optional.";
 
-  var style = findInStylesByName(currentDoc().allCharacterStyles, name);
-  if(!style) {
-    style = currentDoc().characterStyles.add({name: name});
+  if(!arguments || arguments.length > 2) {
+    error(styleErrorMsg);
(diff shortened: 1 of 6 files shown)


b-g:
Hi @trych, many thanks for this! Really great to have these functions! (+ Very beautiful example :) As this quite an important set of functions ... would it not be good to have a few tests for them? (I can merge straight away, but would feel more comfortable with minimum test coverage, or?)

trych:
Good call. I will add some testing for this, so don't merge yet.
The testing suite needs some love (= some more tests) anyways at some point, but that would be a different issue.

About the example script(s): For the functions I am adding/improving right now, I will mostly create example scripts that demonstrate all the different options for the functions. So at the moment the results are not the most beautiful, but I wanted to just get out an example script first. Later people are super welcome to step in and change the script to produce something nicer. :wink:

trych:
I added testing for the new functions now, so the PR can be reviewed. 🔍

b-g:
@trych Many thanks!! Looks good! Merged! (thanks for the linting of the test file)

PR #149, opened 2016-12-21
Move from # to @

fabianmoronzirfas:
This has been bothering me for a long time.  
Changed the # to @ in main.js and all test.  

All test passing:  

```text
58 test cases run
289 assertions run
289 passed
0 failed
```

Test files could use some linting. Examples still need to be changed.  
I also noticed that some of the benchmarking tests where linking to the "release" location of Basil. I changed that so they point to the local version.


b-g:
@fabianmoronzirfas Many thanks for this! Merged!

PR #150, opened 2016-12-21
Replace # with // @ in example scripts and tutorials

trych:
Had this sitting in my local repo for a while, but had not made a PR. But with Fabian's PR earlier today, this fits quite well now (although I just had to redo due to some mess up). 😉

This changes all the `#include`, `#includepath` and `#targetengine` to `// @include`, `// @includepath` and `// @targetengine` in the example scripts and in the tutorials folder.

The standard basil path (as we have to pick _some_ path for the example scripts) gets changed from `~/Documents/basiljs/bundle/basil.js` to `~/Documents/basiljs/basil.js` as we decided [here](https://github.com/basiljs/basil.js/pull/93#issuecomment-247542466).


fabianmoronzirfas:
BÄM! Ship it baby! 🚀

fabianmoronzirfas:
Ähm. One problem is that we really need to update the docs for that.

trych:
Yes, we need to do that before the 2.0 launch anyways. But I think this should be merged now, as we can then move forward and improve and lint the example scripts. The documentation of the installation process is not part of the github repo at all at the moment, if I am not mistaken. So this should be documented when we set up the website.

fabianmoronzirfas:
We should at least:  
- Change the install scripts
- and this guide http://basiljs.ch/tutorials/installation-and-getting-started/. Anybody who starts now and has no experience will run into errors and tell us "it's not working!"

trych:
Yes, this should absolutely be done before releasing 2.0.

Yet I don't see, why this should stop us from merging the updated example scripts now. We are not releasing anything now, so there won't be any confusion for anyone who downloads the current basil version from the official basil site. On the other site for us internally, it would be very helpful. I have literally changed `// @` back to `#` for new example scripts before merging new features recently and it would be nice, if we could just go with `// @` from now on.

fabianmoronzirfas:
Yes you are right. I didn't think of the develop vs master branch

fabianmoronzirfas:
Merged! Awesome

ffd8:
Hey I've been in (basil.js!) class.. but just a quick question/warning: shouldn't such a change be delayed until 2.0?? This is changing something pretty major in the write-up/examples/etc... I know it brings these linting benefits, but will also create weird 'this way or that way' confusion. From 2.0 – it will be easy to clarify to users what has changed.

trych:
Hi @ffd8,
yes, but this is in the develop branch, where we _are_ preparing the 2.0 release right now, so naturally we have to do the changes before the 2.0 release, i.e. now. I mean we want to do all the linting _before_ we do the release. And since it is the develop version it should not bring any confusion to anybody, as it is not released until 2.0 comes out.

fabianmoronzirfas:
@ffd8 you have the same confusion as I had. This is only in develop not in master. Nobody will have these changes until we merge it. If we change the page and the getting started tutorial we can also merge this before moving to 2.0 (IMO)

ffd8:
@trych @fabianmoronzirfas – aha, ok. Email's from git didn't mention the branch... loaded the page and read too quickly to see the 'dev'. thanks!

PR #152, opened 2016-12-22
linting all test files

fabianmoronzirfas:
These are the linted test files. All tests are passing as expected.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "basiljs",
-  "version": "1.0.10",
+  "version": "1.1.0",
```
fabianmoronzirfas: So we stay at 1.1.0 😄  #153

b-g:
@fabianmoronzirfas Many thanks for this! Great to so this nice and tidy!! Merged into dev

PR #154, opened 2016-12-23
docs(changelog): added latest changes to changelog

fabianmoronzirfas:
Added changelog infos. Is it okay like this for you? If so I'll go ahead and merge right away.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -3,7 +3,7 @@
 - removed
 
 ..-  --.- ..- -.... -..-- .-..-. -.-..---.-.-....--.-- -....-.... -..-- .-
-basil.js 1.1.0 22 December 2016
+basil.js x.x.x 22 December 2016
 
 * Changed all include/includepath/targetengine statements from # to //@
   This allows linting with eslint of all files
@@ -12,7 +12,7 @@ basil.js 1.1.0 22 December 2016
 The user is allowed to have his files whereever he wants.  
 
 ..-  --.- ..- -.... -..-- .-..-. -.-..---.-.-....--.-- -....-.... -..-- .-
-basil.js 1.1.0 25 November 2016
+basil.js x.x.x 25 November 2016
 
 + Added b.applyParagraphStyle(), b.applyCharacterStyle(), b.applyObjectStyle(),
   set the style of a text object/page item.
--- a/changelog.txt
+++ b/changelog.txt
@@ -3,16 +3,7 @@
 - removed
 
 ..-  --.- ..- -.... -..-- .-..-. -.-..---.-.-....--.-- -....-.... -..-- .-
-basil.js x.x.x 22 December 2016
-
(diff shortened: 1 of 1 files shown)


fabianmoronzirfas:
realted to #154

b-g:
Looks good! But we won't communicating this release right? This is just for us to keep track of the changes, or?

trych:
I would live the version at `basil.js x.x.x` for now and only fill in when we release. The other changes I would all collect in _one_ list.

fabianmoronzirfas:
@b-g as long as it is in the develop branch anybody will see it. This is just for collecting the changes and not have to rebuild this on a release from git commits.  
@trych I added this because if the list grows long we will have to scroll down down down to see where this is changed but I can keep it at x.x.x as well

trych:
What I want to say is, we should expand this list

```js
basil.js x.x.x

+ Added b.applyParagraphStyle(), b.applyCharacterStyle(), b.applyObjectStyle(),
  set the style of a text object/page item.
  see examples/typography/styles.jsx

* b.paragraphStyle(), b.characterStyle() and b.objectStyle() are expanded:
  They can now get an applied style from a text object/page item and can
  make use of an optional object with property name/value pairs to set
  the properties of the style.
* Improved the speed of b.clear() considerably
```

to this list

```js
basil.js x.x.x

+ Added b.applyParagraphStyle(), b.applyCharacterStyle(), b.applyObjectStyle(),
  set the style of a text object/page item.
  see examples/typography/styles.jsx

* b.paragraphStyle(), b.characterStyle() and b.objectStyle() are expanded:
  They can now get an applied style from a text object/page item and can
  make use of an optional object with property name/value pairs to set
  the properties of the style.
* Improved the speed of b.clear() considerably
* Changed all include/includepath/targetengine statements from # to //@
  This allows linting with eslint of all files
* Linted all the test files
* Changed the location of the basil.js library from `~/Documents/basiljs/bundle/basil.js` to `~/Documents/basiljs/basil.js` in all the example files. 
  The user is allowed to have his files whereever he wants. 
```

and just keep adding (feel free to change the order to place most important features first). But in the end the user needs to see all the features that are new to 2.0.0 in one list and not separated by dates.

trych:
> This is just for collecting the changes and not have to rebuild this on a release from git commits.

Yes, that's generally a good idea and what we should do.

> I added this because if the list grows long we will have to scroll down down down to see where this is changed but I can keep it at x.x.x as well

I would to it x.x.x. The other way kind of implies that there have been several releases (that confusingly have the same version numbers). The changelogs of software I usually see are mostly sorted by importance of features, I have never seen one sorted by date where the feature was pulled in.

trych:
Cool, looks good now. Eventually we should sort it by importance of feature, but we could do that in the end. Important for now is just to collect everything, so we don't have this confusion what we added when the time of release has come.

Looks good for me, feel free to merge.

Issue #155, opened 2016-12-23
Do we still recommend the '~/Documents/basiljs/user' folder?


fabianmoronzirfas:
As the title says. I would love to go ahead and change the installers so the question is:  

Do we still need this folder? I don't think so. Basil is now portable and the user should be able to place all his files wherever he wants them to be. Enforcing the user folder would also mean that all changes made by @trych in PR #150 are not right 🤓

b-g:
I'm happy to change this recommendation with 2.0 ... my blueprint would be Processing e.g.

```
~/Documents/Processing3/examples
~/Documents/Processing3/libraries
~/Documents/Processing3/modes
~/Documents/Processing3/templates
~/Documents/Processing3/tools
```

trych:
What do you mean by installers? The download bundle? Or the docs?

What is your idea of the workflow, where would you put the basil.js file? I personally will still one central spot (and for this `~Documents/basiljs` makes sense) and not have a copy of basil.js for each script that I write.

No, we don't need really need the folder anymore, however I feel it could be helpful to at least suggest one location where you could put basil.js, so inexperienced users have a guide what to do. The problem is, the example scripts need to point to _some_ location, so I thought pointing it to `~Documents/basiljs/basil.js` makes more sense. If there are better alternatives, I am happy to make another PR to change the paths again. :)

b-g:
+1 IMO we should recommend a central location to help inexperienced users:

```
~/Documents/basiljs/basil.js
~/Documents/basiljs/sketches/
~/Documents/basiljs/examples/ ...
~/Documents/basiljs/libraries/ ... where to put e.g. underscore.js
```

fabianmoronzirfas:
I currently not talking about the location of Basil itself. I mean the user folder. If we have Basil at `~/Documents/basiljs/basil.js` there is no room for the user folder. Telling the inexperienced user to just replace a part of that folder will create errors. That is why the `user` folder next to the `bundle` folder was introduced.   

@b-g how to update basil.js in this structure?

@trych The `install.bat` and `install.command` scripts

b-g:
I would tell them to put all their "stuff" into `sketches` and get rid of the `user` folder.

fabianmoronzirfas:
So the update process would be:  

1. Download latest version
2. move `sketches` out of `~/Documents/basiljs`
3. replace the whole folder `~/Documents/basiljs` 
4. move sketches folder back in

---

Or 
1. Download just basil.js file
[...]

PR #157, opened 2017-01-02
.gitignore - Added user folders + minor cleanup

trych:
This adds the user folders `libraries/` and `sketches/` to `.gitignore` as with basil.js's new location the developer's environment is now on the same directory level as the user's environment. (See #155)

+ Some minor cleanup, removal of obsolete paths and making syntax more consistent.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,19 +1,21 @@
 .DS_Store
 basil.sublime-workspace
-examples/image/*.indd
-lib/extendables/log/extendables.log
 *.idlk
-# test files
+
+# Test files
 test/download/data/
 test/download/lib/
 test/download/basil.js
 test/loop/basil.js
 test/loop/lib/
 
+# User directories
+libraries/
+sketches/
+
 # Dependency directories
-node_modules
+node_modules/
 # Logs
-logs
 *.log
 npm-debug.log*
 


fabianmoronzirfas:
Looks good. Currently I suggest an different approach for development.  

On my machine I have the following setup:  

    ~/Documents/basiljs
        .
        ├── bundle
        ├── develop
        └── wiki

This allows me to have the release version side by side with the development version. All test files I write are in the folder `develop/test/*` to link them relativ within the structure.   

For the future versions of Basil (2.*) the .gitignore like this is needed. My fear for this is that we create f\*\*\*ups by committing to the master branch 😞

trych:
> On my machine I have the following setup:

~/Documents/basiljs
    .
    ├── bundle
    ├── develop
    └── wiki

How would you run the example scripts with the new default path `// @include "basiljs/basil.js";// @include "basiljs/basil.js";` in this setup?

It would be nice, if the new `.gitignore` folders could be merged now, so people can make use of them right away. It does not hurt if they are there, even if some developers have a different setup (I guess everybody should just set up his environment as it fits best for them).

fabianmoronzirfas:
> How would you run the example scripts with the new default path (…)
Well. hm. Don't know currently. I never touched the example scripts before :-)

fabianmoronzirfas:
But until we merge the develop branch into master my folder structure is still valid. Or am I missing something?

trych:
> But until we merge the develop branch into master my folder structure is still valid.

Well, it is in a weird state in-between. To do something on the example scripts with the new path (in the develop branch), you would have to change your folder structure. To do anything in the master branch you would have to keep it. If you're not touching the example scripts, you can leave everything as it is. :)

So I guess it still makes sense to add these folders to `.gitignore` in the develop branch for those that already changed the folder structure.

b-g:
@fabianmoronzirfas I would vote to merge this into the dev branch now. I agree with @trych that the new structure should be painless to use for us developers in the dev branch. Please veto or merged the PR straight away. Cheers! :)

PR #162, opened 2017-01-09
Fixed line endings in the remaining broken files

trych:
Ugh. There were *still* some files with broken line endings that did not display correctly in GitHub. This included `template.jsx` which consequently made every derived file from that file broken, which turned out to be quite the annoying issue in my class. 😞

I guess Fabian has a point to write all files from scratch. 😝

As this does not change the functions of the files itself, I will take the freedom to just merge this myself right away.


b-g:
@trych Thanks!!

fabianmoronzirfas:
> I guess Fabian has a point to write all files from scratch. 😝

If I would say "I told you so" I would lie 🤓

PR #165, opened 2017-02-06
b.isText() fix to correctly identify multiple text objects

trych:
During working on another function I noticed that instances of multiple text objects (`Characters`, `Words`, `Lines`) are not recognized by `b.isText()`.

This means that until now this happened:

```js
var tf = b.text(b.LOREM, 0, 0, 200, 200);
var w = b.words(tf);
b.println(w); // --> [object Words]
b.println(b.isText(w)); // --> false
```

This PR fixes this, so collections of multiple Word objects evaluate to true with `b.text()`.

The reason this needs to be evaluated via `obj.constructor.name` instead of `obj instanceof` is that `obj instanceof Characters` etc. throws a reference error if there was not yet an `[object Characters]` defined during script execution.

The order of the checks was slightly altered to improve performance.

Tests for all possible cases were added. It also passes all other tests.

Please review! :)

### DIFF
--- a/basil.js
+++ b/basil.js
@@ -3083,8 +3083,7 @@ var isText = pub.isText = function(obj) {
          obj.constructor.name === "Lines" ||
          obj.constructor.name === "TextStyleRanges" ||
          obj.constructor.name === "Paragraphs" ||
-         obj.constructor.name === "TextColumns" ||
-         obj.constructor.name === "Texts";
+         obj.constructor.name === "TextColumns";
 };
 
 
--- a/basil.js
+++ b/basil.js
@@ -3070,12 +3070,20 @@ var isString = pub.isString = function(str) {
 var isText = pub.isText = function(obj) {
   return obj instanceof Character ||
          obj instanceof InsertionPoint ||
+         obj instanceof Word ||
          obj instanceof Line ||
+         obj instanceof TextStyleRange ||
          obj instanceof Paragraph ||
          obj instanceof TextColumn ||
-         obj instanceof TextStyleRange ||
          obj instanceof Text ||
-         obj instanceof Word;
+         obj instanceof Characters ||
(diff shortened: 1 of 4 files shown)


basil.js
```diff
@@ -3068,14 +3068,23 @@ var isString = pub.isString = function(str) {
  * @return {Boolean} returns true if this is the case
  */
 var isText = pub.isText = function(obj) {
```
fabianmoronzirfas: Looks good to me.

test/data-tests.jsx
```diff
@@ -108,6 +112,36 @@ b.test("DataTests", {
...
+    assert(b.isText(tf.textStyleRanges));
+    assert(b.isText(tf.textStyleRanges.firstItem()));
+
+    assert(b.isText(b.paragraphs(tf)));
+    assert(b.isText(b.paragraphs(tf).firstItem()));
+
+    assert(b.isText(tf.textColumns));
+    assert(b.isText(tf.textColumns.firstItem()));
+
+    // anyone know if and how an [object Texts] can be created?
```
fabianmoronzirfas: I think there is no way to create a `Text` object. It seems to be a class all other text thingies are derived from. The `Texts` collection has no `add` method and is read only.

trych: Hm, you can create a `Text` object by using `itemByRange()` (as I did in line 144). These text objects all don't have add methods, yet you can create (or rather retrieve) most of them. However, I still see no way to create a `Texts` object. If you don't know of a way either, I think I should take it out from the checks again. Otherwise it would mean an unnecessary check for every checked object.
So, please don't merge yet.

trych: Alright, removed this, should be good to go now.

b-g:
@trych Thanks for the bugfix!👍🔧  + merged

PR #169, opened 2017-02-12
documentation node_module and api.json creation

fabianmoronzirfas:
This is just for creating the docs. No changes to the library introduced.


b-g:
@fabianmoronzirfas
Wow, very cool! 

The docs generate with `npm run docs-html` look already very promising!

questions/comments:
* I guess it is on purpose that `npm run watch` also creates all the time `api.json`, may I ask why this makes sense?
* How could the old custom doc categories be transferred to the new docs worksflow. It seems grouping is possible e.g. `Vector` is a single list item in the left column.

fabianmoronzirfas:
Sorry. I did not want to add this to the watch task. Just to the bundling (but the watch task uses bundling 🤔) .  

I actually don't wan't to use the generated HTML just the .json. See https://basiljs.github.io/ I'm trying to make sense of the generated JSON. It is quite large and has lot's of infos that we might not need.

b-g:
Ok + Merged + Thanks!

The generated HTML would be fine fore me (if it comes with better categories/grouping) ... Do you really want to put so much effort into it, just for the sake of having it look more "basil website style" like?

fabianmoronzirfas:
The problem I see is adding groups to the existing html layout will give me the same amount of headache as writing it myself from scratch. I already have [a way to extract the data we need](https://github.com/basiljs/basiljs.github.io/blob/master/.bin/json-transform.js) from the api.json. There are some parts currently still missing. For me it feels safer to write this using liquid then trying to transform it using some JS magic.

fabianmoronzirfas:
Also. Yes. Some nice basil style will be there to 😉

b-g:
Ok + Sure! Nobody will say no! :)

trych:
I remember in the Skype call we decided to have it structured like the p5.js docs, which means one overview page with all functions and then a small sub-page for each single function. Are you still planning for this?

Would be nice, because especially documentation for stuff like `b.color()`, errr, `color()` will be quite detailed as it has many different options (compare the [p5.js color doc](http://p5js.org/reference/#/p5/color) for example).

fabianmoronzirfas:
Could be done. I could generate sub pages from the data we have.  Should we take these conversations over into the pages repo?

PR #179, opened 2017-02-23
. after noiseDetail created null entry in method

fabianmoronzirfas:
minor fix. `.` created problems.


fabianmoronzirfas:
I'm so free to merge this right away.

PR #188, opened 2017-05-03
Bundled basil.js and added missing b.swatch() mention to changelog

trych:
Hey guys,
this adds the missing stuff from the last b.swatch() commit. It adds the basil.js bundle and mentions `b.swatch()` in `changelog.txt`. Could somebody merge this quickly? I committed this in preparation of a larger PR that I would like to create soon. :)

Thanks!

P.S.: Not sure, why git messes up the diffs of the old basil.js and the new one, but the new one is just the bundled version of the source files, so it should be fine.


ffd8:
@trych thanks for putting those items together!

PR #190, opened 2017-05-03
Inspect

### LINKED ISSUE #142: extend b.inspect
inspect is buggy and should be extended.

```js
function prop (f) {
  $.writeln (f.reflect.name);
  var props = f.reflect.properties;
  var array = [];
  for (var i = 0; i < props.length; i++)
  try {array.push (props[i].name + ": " + f[props[i].name])} catch (_){} array.sort ();
  $.writeln (array.join ("\r"));
}

function meth (m) {
  var props = m.reflect.methods.sort(); $.writeln ("\rMethods");
  for (var i = 0; i < props.length; i++)
  $.writeln (props[i].name);
}
```

Taken from Peter Kahrels ScriptUI Handbook http://www.kahrel.plus.com/indesign/scriptui.html

trych:
Whew, this took a while …
but now I a quite happy with the results. 😎

# The all new `b.inspect()`

`b.inspect()` has been completely reworked to get rid of its flaws and improve it in general. I have been using this new version for a while now in the production of some scripts and it turns out to be extremely useful (and teaches you a lot about properties and methods 😉 ). The following improvements, fixes and changes have been made:

## Speed

The new implementation is a lot faster than the old one. For complex objects this means it can be more than 100x faster (that's 100 _times_, not 100 percent). This is achieved by avoiding a lot of recursion, as well as collecting all the results in a virtual string first, before it gets printed once in the end. I created a [benchmark file](https://github.com/trych/basil.js/blob/inspect/test/benchmarks/oldInspectVsNewInspect.jsx) that compares both functions, here are some results:

```
-----------------------------------------
Inspecting a Boolean (100 loops)
(old): 0.514 s
(new): 0.306 s
Speed up factor: 1.6797385620915
-----------------------------------------
-----------------------------------------
Inspecting Numbers (121 loops)
(old): 0.349 s
(new): 0.381 s
Speed up factor: 0.91601049868766
-----------------------------------------
-----------------------------------------
Inspecting a String (100 loops)
(old): 0.32 s
(new): 0.508 s
Speed up factor: 0.62992125984252
-----------------------------------------
-----------------------------------------
Inspecting a Custom Object @ Level 1 (10 loops)
(old): 0.262 s
(new): 0.036 s
Speed up factor: 7.27777777777778
-----------------------------------------
-----------------------------------------
Inspecting a Custom Object @ Level 2 (10 loops)
(old): 0.655 s
(new): 0.065 s
Speed up factor: 10.0769230769231
-----------------------------------------
-----------------------------------------
Inspecting an Array @ Level 1 (10 loops)
(old): 0.273 s
(new): 0.034 s
Speed up factor: 8.02941176470588
-----------------------------------------
-----------------------------------------
Inspecting an Array @ Level 2 (10 loops)
(old): 0.684 s
(new): 0.05 s
Speed up factor: 13.68
-----------------------------------------
-----------------------------------------
Inspecting a Color @ Level 1
(old): 0.042 s
(new): 0.006 s
Speed up factor: 7
-----------------------------------------
-----------------------------------------
Inspecting a Color @ Level 2
(old): 1.795 s
(new): 0.182 s
Speed up factor: 9.86263736263736
-----------------------------------------
-----------------------------------------
Inspecting a Color @ Level 3
(old): 3.479 s
(new): 1.57 s
Speed up factor: 19.1153846153846
-----------------------------------------
-----------------------------------------
Inspecting a Rectangle @ Level 1
(old): 0.115 s
(new): 0.063 s
Speed up factor: 1.82539682539683
-----------------------------------------
-----------------------------------------
Inspecting a Rectangle @ Level 2
(old): 0.706 s
(new): 0.234 s
Speed up factor: 3.01709401709402
-----------------------------------------
-----------------------------------------
Inspecting a Rectangle @ Level 3
(old): 4.573 s
(new): 2.511 s
Speed up factor: 19.542735042735
-----------------------------------------
-----------------------------------------
Inspecting App @ Level 1
(old): 74.464 s
(new): 0.243 s
Speed up factor: 306.43621399177
-----------------------------------------
-----------------------------------------
Inspecting $ @ Level 1
(old): 0.174 s
(new): 0.011 s
Speed up factor: 15.8181818181818
-----------------------------------------
```
The exact speed-up depends a lot on the type of object you are passing and on the `maxLevel` etc. The new function was only slower for inspecting simple strings and numbers and probably only because I chose to show their properties and methods during the tests. If you want to use the file to compare, you need to checkout [this commit](https://github.com/trych/basil.js/commit/9d54b16581defff3135d6bbfcec38aa25cbab262) that contains both the old and the new method next to each other.

## Presentation
The presentation of the results has been cleaned up a lot, so it should be much easier to read and understand what you are inspecting, especially for more complex objects. A tree style diagram is "drawn" to the console to present the results. I spent quite a while to try to achieve a look that is easily legible, I hope you guys like the result. The `PROPERTIES` and `METHODS` sections are highlighted for easy distinguishing and for each recursion level down you will see a new branch.

Example output of a simple nested array `["I", "love", "Basil", [true, false], 23, 5, b.color(255, 0, 0)]:

```js
[I, love, Basil, true,false, 23, 5, [object Color]] (Array)
|
|   PROPERTIES
|-- length: 7
|-- 0: I
|-- 1: love
|-- 2: Basil
|-- 3: Array (2)
|   |
|   |   PROPERTIES
|   |-- length: 2
|   |-- 0: true
|   |-- 1: false
|   
|-- 4: 23
|-- 5: 5
|-- 6: [object Color]
|   |
|   |   PROPERTIES
|   |-- colorValue: Array (3)
|   |-- eventListeners: [object EventListeners]
|   |-- events: [object Events]
|   |-- id: 215
|   |-- index: 10
|   |-- isValid: true
|   |-- isValid: true
|   |-- label: 
|   |-- model: PROCESS
|   |-- name: R=255 G=0 B=0
|   |-- parent: [object Document]
|   |-- properties: [object Object]
|   |-- space: RGB
|
|   METHODS
|-- concat()
|-- filter()
|-- join()
|-- map()
|-- pop()
|-- push()
|-- reverse()
|-- shift()
|-- slice()
|-- sort()
|-- splice()
|-- toLocaleString()
|-- toSource()
|-- toString()
|-- unshift()
```

The properties and methods are sorted alphabetically to find stuff easier. With complex objects the old method was almost unusable, the new method makes it much cleaner:

inspectnewold

## Options

The new method has a `settings` parameter, which has parameters to control the result of `b.inspect()`. The parameters are (quoting from the new documentation):

- `showProps`: Show or hide properties. Default: `true`
- `showValues`: Show or hide values. Default: `true`
- `showMethods`: Show or hide methods. Default: `false`
- `maxLevel`: Chooses how many levels of properties should be inspected recursively. Default: `1`
- `propList`: Allows to pass an array of property names to show. If propList is not set all properties will be shown. Default: `[]` (no propList)

If no settings object is set, the default values will be used. The `propList` parameter is useful when I want to only observe a few specific properties of a complex object. For example, observe only the `geometricBounds` and `strokeWeight` values of a rect: `b.inspect(myRect, {maxLevel: 2, propList: ["geometricBounds", "strokeWeight"]};`.

Btw. I set the default for `showMethods` to `false` now, as in my tests I got the impression that it is a lot more confusing for beginners to have these methods around (as beginners probably don't know what methods even are). If they learn about methods at some point, they can always choose to show them. Is this okay for you guys?

## Reliability

The function is a lot more robust now and can handle all sorts of data types. The old version for example could not handle `null` or `undefined` values, which is something that could quite possibly be the value of a variable that is passed. Inspecting `null` or `undefined` works now. The new function also correctly shows the values of Enumerators, which might look like this:

```js
|-- strokeAlignment: CENTER_ALIGNMENT
```
while the old function would only state

```js
strokeAlignment: Enumerator
```

---
---

That's it. The code is quite complex, since a lot of exceptions need to be handled.

This closes #142. Please review. I think for reviewing you probably need to ~~load the branch~~ (well you actually don't need to load a branch anymore, you can just grab the bundled `basil.js` file from my branch – the perks of having basil in a single file 😉  ) and actually make use of the new `b.inspect()` function to test if it works for you.

@fabianmoronzirfas Could you also please carefully review the documentation of the improved method? I tried to follow the new jsDoc guidelines, I was not quite sure how to handle the settings object. I gave three examples to make it easy to understand how to use it with and without `settings` object and with the `propList`.


ffd8:
Excited to try this! Ran into plenty of the annoying 'null' bug when inspecting something like the NYTimes API's in our recent class... will try to give feedback soon.

src/includes/environment.js
```diff
@@ -749,61 +749,191 @@ pub.bleeds = function(top, right, bottom, left) {
...
+ * b.inspect(b.page(), {showMethods: true, maxLevel: 2})
+ *
+ * @example
+ * // inspecting an ellipse, listing only the properties "geometricBounds" and "strokeWeight"
+ * var myEllipse = b.ellipse(0, 0, 10, 10);
+ * b.inspect(myEllipse, {maxLevel: 2, propList: ["geometricBounds, strokeWeight"]});
  */
-pub.inspect = function(obj, maxlevel, level, propname) {
-  checkNull(obj);
+pub.inspect = function (obj, settings, level, branchArray, branchEnd) {
```
fabianmoronzirfas: What are `level`, `branchArray` and `branchEnd` for? And why are they undocumented?

src/includes/environment.js
```diff
@@ -749,61 +749,191 @@ pub.bleeds = function(top, right, bottom, left) {
...
+ * <code>maxLevel<code>: Chooses how many levels of properties should be inspected recursively. Default: <code>1<code>
+ * <code>propList<code>: Allows to pass an array of property names to show. If propList is not set all properties will be shown. Default: <code>[]<code> (no propList)
+ * If no settings object is set, the default values will be used.
  *
  * @cat Output
  * @method inspect
- * @param  {Object} obj The Object to be inspected.
- * @param  {Number} maxlevel Recursion limit, default maxlevel = 2.
+ * @param  {Object} obj An object or any other data item to be inspected.
+ * @param  {Object} settings A settings object to control the function's behavior.
```
fabianmoronzirfas: The way to here would be like [this](http://usejsdoc.org/tags-param.html#parameters-with-properties)  

```js
/**
 * My fancy object paramter.
 * @param {Object} obj - The description of the object.
 * @param {string} obj.name - The description of the objects properties.
 * @param {string} obj.andwhatnot - and even more description.
 */
```

basil.js
```diff
@@ -2223,61 +2223,191 @@ pub.bleeds = function(top, right, bottom, left) {
...
  *
  * @cat Output
  * @method inspect
- * @param  {Object} obj The Object to be inspected.
- * @param  {Number} maxlevel Recursion limit, default maxlevel = 2.
+ * @param  {Object} obj An object or any other data item to be inspected.
+ * @param  {Object} settings A settings object to control the function's behavior.
+ *
+ * @example
+ * // inspecting a string
```
fabianmoronzirfas: the comment might go into a `<caption>tag</caption>` directly after the `@example` tag. Same goes for the other examples. Would be better to highlight the example code.

fabianmoronzirfas: ups. The review should have been done on the `environment.js` not on the bundle.

src/includes/environment.js
```diff
@@ -749,61 +749,191 @@ pub.bleeds = function(top, right, bottom, left) {
...
+ *
+ * @example
+ * // inspecting a string
```
fabianmoronzirfas: the comment might go into a `<caption>tag</caption>` directly after the `@example` tag. Same goes for the other examples. Would be better to highlight the example code.

fabianmoronzirfas:
To things I want to address.

1. In General: Would be great if you start working on branches directly in the basil.js repo instead of your fork. This makes checking out new branches for inspection/review easier. (Currently I keep a copy of your repo around as well)
2. About `inspect`: I would vote for making the `showMethods` by default `true`. This should not be hidden from the user. (I'm not sure if we already hat that discussion…)
3. **Awesome work.** The tree view looks already great. Do you know the [`tree`](http://mama.indstate.edu/users/ice/tree/) command (brew install tree). It gives you a listing of files in folders and their hierarchy. e.g.  

```txt
├── doc
├── extras
├── lib
├── node_modules
├── scripts
├── src
└── test
```
 
What I really like is that the last element is showing with the `└──` that it is the last. Might be cool for `b.inspect` as well.

b-g:
@trych I just can add a "wow"!! This is really mega!

src/includes/environment.js
trych: Help me out here, so like this:
```js
/**
 * ...
 * @cat Output
 * @method inspect
 * @param  {Object} obj An object or any other data item to be inspected.
 * @param  {Object} settings A settings object to control the function's behavior.
 * @param  {String} settings.showProps Show or hide properties. Default: <code>true<code>
 * @param  {String} settings.showValues Show or hide values. Default: <code>true<code>
 * etc.
 */
```

?

src/includes/environment.js
trych: They are used internally only for the recursive call of the function, so that's why they should not be exposed to the user (as they are not supposed to use these parameters). To help with reviewing though:

`level` holds the current level (top level being `0`)
`branchArray` is an array of booleans that holds the info how the branches need to be drawn in the current line. Something like `[true, true, true]` renders to

```
|   |   |   
```
 while something like `[true, false, true]` renders to
```
|       |   
```
`branchEnd` is a boolean that holds the information if the parent property is at the very end of a branch. This is needed for correctly drawing the tree.

basil.js
trych: Cool, will change that. Coming up in a new PR soon.

src/includes/environment.js
trych: 👍

trych:
@fabianmoronzirfas 

> Would be great if you start working on branches directly in the basil.js repo instead of your fork.

I knew that day would come. 😉 Haha, no you're right, I will make the switch soon, just need to figure out how to transfer unfinished and experimental branches first.

> About inspect: I would vote for making the showMethods by default true. This should not be hidden from the user.

What do the others think? @b-g @ffd8 Personally I think it's less confusing for beginners, if it not shown by default, but if you all want it shown by default, no problem.

> Do you know the tree command (brew install tree).

Yes, I actually know it from some of your posts here on GitHub, that's also where the inspiration came from.

> What I really like is that the last element is showing with the └── that it is the last. Might be cool for b.inspect as well.

I totally agree, and I also tried this at first, the problem is, that the `Run in InDesign` package in Sublime (that I use exclusively) can only render ASCII characters at the moment. That means something like

```js
|-- 4: 23
|-- 5: 5
|-- 6: [object Color]
|   |
|   |   PROPERTIES
|   |-- colorValue: Array (3)
|   |-- eventListeners: [object EventListeners]
|   |-- events: [object Events]
|   |-- id: 215
|   |-- index: 10
|   |-- isValid: true
|   |-- isValid: true
|   |-- label: 
|   |-- model: PROCESS
|   |-- name: R=255 G=0 B=0
|   |-- parent: [object Document]
|   |-- properties: [object Object]
|   |-- space: RGB
|
|   METHODS
|-- concat()
|-- filter()
```

using these fancy `└──` characters would look something like this right now:

```js
??? 4: 23
??? 5: 5
??? 6: [object Color]
?   ?
?   ?   PROPERTIES
?   ??? colorValue: Array (3)
?   ??? eventListeners: [object EventListeners]
?   ??? events: [object Events]
?   ??? id: 215
?   ??? index: 10
?   ??? isValid: true
?   ??? isValid: true
?   ??? label: 
?   ??? model: PROCESS
?   ??? name: R=255 G=0 B=0
?   ??? parent: [object Document]
?   ??? properties: [object Object]
?   ??? space: RGB
?   METHODS
??? concat()
??? filter()
```

So that does not work obviously, so I tried to find the cleanest way using ASCII characters only instead. What I suggest we do: Merge it in its current (ASCII only) implementation for now, file an issue at the `Run in InDesign` repo to get the rendering of non-ASCII characters. And if this is done, then we could easily switch to proper `└──` characters. Should be a simple replacement of the strings, plus some small tweak for the last item of a branch, nothing dramatic. What do you think?

src/includes/environment.js
fabianmoronzirfas: Yep. thats right.

src/includes/environment.js
fabianmoronzirfas: okidoki.

src/includes/environment.js
trych: Hang on, but it probably should say `@param {Boolean} settings.showValues`, right? Not `{String}`. It probably should always state the type of the sub-property, correct?

basil.js
trych: Oh, and I messed up all the closing tags. HTML noob that I am … ;)

src/includes/environment.js
trych: Ok, updated the PR. Could you have another look?

src/includes/environment.js
```diff
@@ -750,28 +750,33 @@ pub.bleeds = function(top, right, bottom, left) {
...
  * @param  {Object} obj An object or any other data item to be inspected.
- * @param  {Object} settings A settings object to control the function's behavior.
+ * @param  {Object} [settings] A settings object to control the function's behavior.
+ * @param  {Boolean} [settings.showProps] Show or hide properties. Default: <code>true</code>
+ * @param  {Boolean} [settings.showValues] Show or hide values. Default: <code>true</code>
+ * @param  {Boolean} [settings.showMethods] Show or hide methods. Default: <code>false</code>
+ * @param  {Number} [settings.maxLevel] How many levels of properties should be inspected recursively. Default: <code>1</code>
+ * @param  {Array} [settings.propList] Array of properties to show. Default: <code>[]</code> (no propList)
  *
  * @example
```
fabianmoronzirfas: the caption needs to be in the line of the `@example` tag. Should be

    * @example <caption>Inspecting a string</caption>

basil.js
fabianmoronzirfas: hehe. Didn't see that. Great you did.

trych:
@fabianmoronzirfas Moved the captions behind the examples tag.

Did anybody have any chance yet to test this out?

fabianmoronzirfas:
I had a quick look but no extensive testing. Caught up in work around here. Sorry. But I think you did a lot of testing with it. It looks way better then the inspect before and I'm sure you tested it on any kind of object. So I would say go ahead and merge. @ffd8 @b-g What do you say?

trych:
Yes, I did a lot of testing, I think it should work a lot better generally. It's a bit difficult to test every user scenario with this function, so I guess if there are still minor issues, they should become apparent it this function is used over time and people report back. So, yes, if nobody is generally against this function or its implementation, I would also vote to merge it.

@ff8d @b-g , do you agree?

And before I merge it, I still would like to know your opinions on the default of `showMethods`. Fabian wants it to show by default, I tend to not show it by default, what do the others think?

trych:
Btw. is GitHub acting weird today or is this something strange on my end? None of the text formatting buttons work for me and I cannot edit my posts, because the edit button does not work for me, neither the emoji button. Did they change something recently?

fabianmoronzirfas:
**I** _dont_
# know 

### if 

> something

 `is worng`
[with GitHub](https://example.com).

- Works

1. good for me.

ffd8:
I was also having weird issues today with github - but only in an older version of safari = aaargh.

@trych finally sat down to test it (and agree about making it a branch in the main repo for easy testing- but now I've cloned you! so.. yeah)

I gave it a test from the standpoint of our recent in class project of NYTimes JSON API. The previous inspect failed whenever a null was reached, making it pretty useless. This new one is a major major improvement! But a few questions/suggestions:

- I'd drop the PROPERTIES word all together.. with that line gone, it will be so much cleaner to read. I'd keep that blank space as you have it from the parent (so removing PROPERTIES and thus, everything moves up one line).

- maxLevel is great! but it caused some weird issues while looking at a JSON that was quite simple.. it kept doing a 'reflect' that I don't understand. ie. here's an example JSON URL to try loading (just have to make a quick NYTimes API account = 1min)

```
  var apiKey = "insert_API_key";
  var apidate = "20170417";
  var nytURL = "http://api.nytimes.com/svc/community/v3/user-content/by-date.json?api-key="+apiKey+"&date="+apidate;
  var nytJSON = b.loadString(nytURL); 
  var nyt = b.JSON.decode(nytJSON); // decode json to object 
  b.inspect(nyt, {maxLevel:3});
```
it constantly produced this output in the inspect:
```
 PROPERTIES
|   |-- reflect: [object Reflection]
|   |   |
|   |   |   PROPERTIES
|   |   |-- description: 
|   |   |-- help: 
|   |   |-- methods: Array (9)
|   |   |-- name: Object
|   |   |-- properties: Array (5)
|   |   |-- sampleCode: null
|   |   |-- sampleFile: null
|   |   |-- staticMethods: Array (1)
|   |   |-- staticProperties: Array (3)
```

- showValues sounded like a great way to see the structure without the actual contents getting in the way, however when setting it to false, it ignored my setting of `maxLevels:3` – not sure if that's a bug or intentional, since it is ignoring the values... the question is if it could/should respect the request of values? or make an exception for the value being another object or array and going further? ignoring any value that isn't such a thing?

Just inspected a textframe = it's sooo beautiful! Really great work!

trych:
@ffd8 Thanks for the great feedback!

(Yes, I also now had problems in an old Safari environment. That means they changed something overnight without announcing and breaking a lot of stuff like they did recently when an update broke all Markdown headlines ... great. 😞 )

- Dropping the `PROPERTIES` headline.
I personally like it better with the headline, but we can also get rid of it. @fabianmoronzirfas @b-g Any preferences on your side?

- reflect
`reflect` is a property that every object has, even if you create a new custom object with `var myObject = {};`. However, since it is a "meta property" describing only the other properties of the object (which we are doing anyways with `b.inspect()`) it would probably be best to skip the reflect property altogether (as it is currently already done with the properties `__proto__`, `__count__`, `__class__`). I stumbled across this myself already. If you do a `b.inspect()` command on `b.bounds(myWord)`. You will get a weird list of

```js
|-- bottom: -120
|-- height: 280
|-- left: -400
|-- reflect: [object Reflection]
|-- right: -318
|-- top: -400
|-- width: 82
```

and the reflect property in this list is probably confuse everybody and does very little. If somebody knows about the reflect property and needs it, he will be pro enough to access it manually. So I vote to skip it and not list it.

- `showValues`
> showValues sounded like a great way to see the structure without the actual contents getting in the way, however when setting it to false, it ignored my setting of maxLevels:3

Well, yes, that's because the sub-branches themselves are part of the values (as the values in this case are objects or arrays). So, if you don't show the values, you can technically only stay on the top level. So, if you need to see the structure, I would just leave `showValues` on. :)

b-g:
Hi @trych,
Hi all!

Super nice work!! Very helpful for a lot of use cases. Great! 

Very little feedback form my end, as everything already said. Just:

* +1 Dropping the PROPERTIES headline.
* +1 Dropping the meta property reflect.
* showValues, consider renaming it to something more intuitive e.g. shallow, firstLevelOnly ...

Best,
B

trych:
@b-g Thanks for the feedback.

I think renaming showValues to `shallow` is even more confusing, because you can also have shallow when you do `{maxLevels: 1, showValues: true}`. I wonder, if we should remove `showValues` altogether? I think it is of very little advantage to only show the property names. Not sure, if anybody really needs it, as seeing the values right next to them does never really hurt. What do you guys think?

And @b-g (or @ffd8) pleaaaase a quick opinion on the default of `showMethods`: `true` or `false`? :)

ffd8:
- I vote for `showMethods:false` by default. Creates noise, but available for the very curious. 

- `showValues`, I know what you mean by any object/array within the prop being a value.. buuuut if it was possible to allow an array and object to slide? – it would be super helpful. By default the user has the maxLevels at 1.. so values off = no big deal. The minute one wants to fully inspect an object and know all possible props one can give.. it's important to see what props exist a few levels down (still avoiding the noise of the actual values). I can think of many scenarios where it's helpful. Just re-read your question if it should be ditched.. no way, always nice to have a cluttered/noise-free view of the object. Haha.. now I just have to convince you of the value:object/array's importance to filter. This method would also be oh so helpful for students to debug an object they've created and are trying to learn to control – inspect the path of levels to get to that damn property they want to change, without the LOOONG list and clutter of all values there. Therefore I'd tell students to use the following to learn about an obj:

`b.inspect(obj, {maxLevels:10, showValues:false});` (improved below)

- maxLevels, implies an attempt to find go deeper if it exists.. is it perhaps more common to want to inspect the entire thing? Like what @b-g was suggesting.. by default it's all levels, but can be specified to be limited to `firstLevel` or keep using `maxLevel:1/2/3` to limit it. But default is 'all'? If all were default, then learning about an object would just be:

`b.inspect(obj, {showValues:false});`

and the way above but reduced if 'too much information' was being dumped.
One of the hardest things I saw students struggle with in objects was that path of obj.propA.propB[i].propC

That just made me wonder if the ascii design of the structure could benefit from sneaking a . in there to help make it intuitive for the path.. but I also really like how clean your current solution is.

trych:
> I vote for showMethods:false by default. Creates noise, but available for the very curious.

@b-g What do you think?

> showValues, I know what you mean by any object/array within the prop being a value.. buuuut if it was possible to ...

Hm, in this case, I would maybe rather rename it to something like `treeOnly: true`, because then it is not really about the values being on or off anymore, but about the tree structure itself. What do you think?

> Like what @b-g was suggesting.. by default it's all levels

That's not gonna work: With InDesign objects there is an extreme amount of branches. Imagine: For every textFrame there is always a `.parent` property, which has a `.parent` property again and so on until you reach at some point the app object, that holds all documents and so much stuff, that you could probably branch it a few dozen levels down, which ...

... will take forever (and probably crash). I just gave a simple rect a test with `{maxLevel: 5}` it took (even with the new fast method) almost ten minutes and returned 429033 (!) lines. This will be not very helpful for anybody. 😜 So I think `1` makes total sense as a default and people need to decide themselves how many levels down they want to go.

> That just made me wonder if the ascii design of the structure could benefit from sneaking a . in there to help make it intuitive for the path

I kind of like that idea, only, I don't know how to visually implement that, especially using only ASCII symbols.

Take something like

```js
|   
|-- strokeType: [object StrokeStyle]
|   |
|   |   PROPERTIES
|   |-- eventListeners: [object EventListeners]
|   |-- events: [object Events]
|   |-- id: 23081
|   |-- index: 17
|   |-- isValid: true
|   |-- isValid: true
|   |-- name: Durchgezogen
|   |-- parent: [object Document]
|   |-- properties: [object Object]
|   |-- strokeStyleType: 
|   
|-- strokeWeight: 0
|-- tables: [object Tables]
|   |
|   |   PROPERTIES
|   |-- length: 0
|   
|-- textBoxes: [object TextBoxes]
```

With points it would look like this:

```js
|   
.-- strokeType: [object StrokeStyle]
|   |
|   |   PROPERTIES
|   .-- eventListeners: [object EventListeners]
|   .-- events: [object Events]
|   .-- id: 23081
|   .-- index: 17
|   .-- isValid: true
|   .-- isValid: true
|   .-- name: Durchgezogen
|   .-- parent: [object Document]
|   .-- properties: [object Object]
|   .-- strokeStyleType: 
|   
.-- strokeWeight: 0
.-- tables: [object Tables]
|   |
|   |   PROPERTIES
|   |-- length: 0
|   
.-- textBoxes: [object TextBoxes]
```

which looks clearly worse. Bullet points don't work in ASCII, but I tried them as well, it does not look much better.

We could do something like this and put a `.` before every property, but I am not sure, if this is any better … probably not?

```js
|   
|-- .strokeType: [object StrokeStyle]
|   |
|   |   PROPERTIES
|   |-- .eventListeners: [object EventListeners]
|   |-- .events: [object Events]
|   |-- .id: 23081
|   |-- .index: 17
|   |-- .isValid: true
|   |-- .isValid: true
|   |-- .name: Durchgezogen
|   |-- .parent: [object Document]
|   |-- .properties: [object Object]
|   |-- .strokeStyleType: 
|   
|-- .strokeWeight: 0
|-- .tables: [object Tables]
|   |
|   |   PROPERTIES
|   |-- .length: 0
|   
|-- .textBoxes: [object TextBoxes]
```

fabianmoronzirfas:
My votes 
+1 on dropping PROPERTIES line
+1 on dropping all meta properties like reflect (can be excluded with `if(obj.hasOwnPropert(key))` I think)
-1 on renaming show values. I think keeping the `key:value` wording is good
-1 on removing showValues. I like it a lot 
+1 on `showMethods:true` I really think it's important not to give a preference to props over methods
-1 on all levels. For performance sake and for UX sake. 
-1 on adding a `.` everything will have a period because everything is a child of  `app`

b-g:
Hi guys, just went over the conversation ... happy with @fabianmoronzirfas's votes too! Reads like a good conclusion of the open decisions. 👍 Ready to merge afterwards from my end (once Fabian's list is implemented). Best + have nice evening! B

trych:
I agree with everything except the `showMethods: true`.

I don't think this is about giving preference to props over methods, I think the defaults that we pick should serve best for the most probable use case of the function. So something like `b.inspect(myRect)` should return the most useful result in most of the use cases. I am certain that 99% of the time the function is used to look up values of certain properties. I don't think anybody will really use it that much to look up method names.

And if somebody does, (or if you want to teach students about methods) they can still set the flag to true. But have it true for _all_ default searches when most people don't need to know the method names is not ideal, I think.

b-g:
As I've never used it in class ... I don't have a strong preference on this detail. From my end either way possible. Maybe the rest of the gang can decide on this w/o me?

trych:
Ok, then since @ffd8 and me are against it and @b-g is neutral, I would leave it set to false by default now. @fabianmoronzirfas hope you can live with that. 😉

The other points I agree on and will prepare another PR soon.

fabianmoronzirfas:
Nooooooooooooooooooooooooo.

trych:
I knew it would break your heart. 💔 😢
To make up for it, I will soon march the streets with a home made sign reading: "Methods have feelings, too!" 😜

ffd8:
https://youtu.be/XW1HNWqdVbk?t=58s

ps. yupyup, was totally wrong on levels:all.. indesign objects are way more complex beasts than most API JSON dumps. all sounds good to me, thanks for the discussion + awesome function!

b-g:
> "Methods have feelings, too!"
😂

trych:
Alright, I implemented these changes. Skipping the `PROPERTIES` flag now and skipping the `reflect` property. So, this should be ready to merge.

@fabianmoronzirfas Is there maybe a chance to merge this still tonight _and_ regenerate the reference again as well? I have this workshop tomorrow and would love to use the latest develop version along with your (up to date) docs (don't worry about the code tag bug for now, that's not urgent for me).

b-g:
@trych merged! Mega thanks!

b-g:
But currently don't know how to refresh the API docs under https://basiljs.github.io/ ... Sorry!

trych:
@fabianmoronzirfas If this is quick and easy and manageable for you to do tonight it would be nice, else don't worry and enjoy your weekend instead. :)

fabianmoronzirfas:
as you know there is in bug in the docs https://github.com/basiljs/basiljs.github.io/issues/4

fabianmoronzirfas:
closed with https://github.com/basiljs/basiljs.github.io/commit/4cad901954ad56aa8dfc23af32c97a0c332a41ec

trych:
Great! Thanks a lot. The `b.inspect()` doc is still the old one though, was it generated from an older version?

PR #192, opened 2017-05-13
found a TODO made it a `@todo`

fabianmoronzirfas:
minor change. Only for the beauty of the docs.  

bildschirmfoto 2017-05-13 um 20 57 46

## VS

bildschirmfoto 2017-05-13 um 20 58 22

(Did you know there is a user called `@todo`?)

### DIFF
--- a/src/includes/shape.js
+++ b/src/includes/shape.js
@@ -179,8 +179,7 @@ pub.vertex = function() {
  *
  * @return {GraphicLine|Polygon} The resulting GraphicLine or Polygon object (in InDesign Scripting terms the corresponding type is GraphicLine or Polygon, not Arc).
  *
- * TODO(S)
- * - fix overlapping points bug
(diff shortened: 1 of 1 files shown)


trych:
I think we should move the ToDos out of the docs altogether. I think as soon as we find a Todo we should create an issue here on GitHub. That's a way more advanced system to keep track of ToDos and the user does not get confused by a ToDo in the reference.

fabianmoronzirfas:
Yes we should can we still merge this? So we don't need to search for \@todo and todo and TODO and TODOS?

b-g:
@fabianmoronzirfas Thanks for spotting that one! This is really really old legacy stuff. Yes, lets remove any "todo" stuff in source code. + Merged.

fabianmoronzirfas:
okidoki. I'll put it on my list.

PR #198, opened 2017-06-16
Renamed vertex params from *anchor* to *handle*

trych:
Came across this while creating Sublime snippets for basil.js. The parameters for `b.vertex()` are named incorrectly. The optional parameters for the handles are named `xAnchorLeft` etc., while the term "anchor" actually always refers to the center node of a bezier node, not the handles.

So, I think this should be more correct and more intuitive to understand.

Please have a look, if this change is ok.

### DIFF
--- a/basil.js
+++ b/basil.js
@@ -3774,7 +3774,7 @@ pub.beginShape = function(shapeMode) {
  * the beginShape() and endShape() functions.
  *
  * Use either vertex(x, y) for drawing straight corners or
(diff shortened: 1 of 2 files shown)


fabianmoronzirfas:
For me this is okay. If no one else has objections I would say let's go ahead and merge this one.

b-g:
Thanks for this improvement! +1 for merge.

fabianmoronzirfas:
I'll merge it.

PR #199, opened 2017-07-13
Added possibility to use a page size preset as argument for b.size()

trych:
Hey there,
just had the idea that it would be nice if instead of using values to set the page size with `b.size()`, you could also set the size with the name of a page size preset like `"A4"` or `"Letter"`, etc.

Was easy to implement, so I decided to create a pull request right away. Hope the idea makes sense to you guys as well.

However, before anybody goes ahead to merge this, one question: Should we also allow yet another parameter to set the orientation? With the implementation I just did, it simply keeps the orientation. That means if I have a landscape `A4` document and set it to `iPhone` format via `b.size("iPhone")` I will get a landscape iPhone sized document automatically.

Should we allow an extra parameter like `b.PORTRAIT` and `b.LANDSCAPE` (in which case I would create a new pull request with that implemented)?

So usage would look like this:

`b.size("A3", b.LANDSCAPE);`

### DIFF
--- a/basil.js
+++ b/basil.js
@@ -1494,44 +1494,56 @@ pub.doc = function(doc) {
 /**
  * Sets the size of the current document, if arguments are given.
  * If only one argument is given, both the width and the height are set to this value.
+ * Alternatively a string can be given as the only argument to apply an existing page size preset ("A4", "Letter" etc.).
  * If no argument is given, an object containing the current document's width and height is returned.
  *
  * @cat Document
  * @method size
- * @param  {Number} width The desired width of the current document.
+ * @param  {Number|String} widthOrPageSize The desired width of the current document or the name of a page size preset.
  * @param  {Number} [height] The desired height of the current document. If not provided the width will be used as the height.
  * @return {Object} if no argument is given it returns an object containing the current width and height of the document.
  *
  */
-pub.size = function(width, height) {
+pub.size = function(widthOrPageSize, height) {
   if(app.documents.length === 0) {
     // there are no documents
-    warning("b.size(width, height)", "You have no open document.");
+    warning("b.size()", "You have no open document.");
     return;
   }
   if (arguments.length === 0) {
     // no arguments given
-    // return the curent values
-    // warning('b.size(width, height)', 'no arguments given');
+    // return the current values
     return {width: pub.width, height: pub.height};
   }
 
-  if(arguments.length === 1) {
+  var doc = currentDoc();
+
+  if(arguments.length === 1 && typeof widthOrPageSize === "string") {
+
+    try {
+      doc.documentPreferences.pageSize = widthOrPageSize;
+    } catch (e) {
+      b.error("b.size(), could not find a page size preset named \"" + widthOrPageSize + "\".");
+    }
+    pub.height = doc.documentPreferences.pageHeight;
+    pub.width = doc.documentPreferences.pageWidth;
+    return {width: pub.width, height: pub.height};
+  } else if(arguments.length === 1) {
     // only one argument set the first to the secound
-    height = width;
+    height = widthOrPageSize;
   }
-  var doc = app.documents[0];
-    // set the documents pageHeiht and pageWidth
+  // set the document's pageHeight and pageWidth
   doc.properties = {
     documentPreferences: {
       pageHeight: height,
-      pageWidth: width
+      pageWidth: widthOrPageSize
     }
   };
   // set height and width
   pub.height = height;
-  pub.width = width;
+  pub.width = widthOrPageSize;
 
+  return {width: pub.width, height: pub.height};
 
 };
 
--- a/basil.js
+++ b/basil.js
@@ -330,6 +330,22 @@ pub.BEFORE = LocationOptions.BEFORE;
 pub.AFTER = LocationOptions.AFTER;
 
 /**
+ * Used with b.size() to set the orientation of a given page size to portrait.
+ * @property PORTRAIT {String}
+ * @cat Document
+ * @subcat Page
+ */
+pub.PORTRAIT = PageOrientation.PORTRAIT;
+
+/**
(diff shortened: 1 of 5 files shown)


b-g:
Great! 👍

Yes to `b.PORTRAIT`, `b.LANDSCAPE` and `b.size("A3", b.LANDSCAPE)`; from my end.

basil.js
```diff
@@ -1494,44 +1494,56 @@ pub.doc = function(doc) {
 /**
  * Sets the size of the current document, if arguments are given.
  * If only one argument is given, both the width and the height are set to this value.
+ * Alternatively a string can be given as the only argument to apply an existing page size preset ("A4", "Letter" etc.).
```
fabianmoronzirfas: - What are the possible presets?
- Can we list them in the docs?
- What about localization?

basil.js
```diff
@@ -1494,44 +1494,56 @@ pub.doc = function(doc) {
...
 
-  if(arguments.length === 1) {
+  var doc = currentDoc();
+
+  if(arguments.length === 1 && typeof widthOrPageSize === "string") {
+
+    try {
+      doc.documentPreferences.pageSize = widthOrPageSize;
+    } catch (e) {
+      b.error("b.size(), could not find a page size preset named \"" + widthOrPageSize + "\".");
```
fabianmoronzirfas: Is there a way to get all the existing preset names and list them here?

fabianmoronzirfas: When running this code:  

```js
var presets = app.documentPresets;
for (var i = presets.length - 1; i >= 0; i--) {
  $.writeln(presets[i].name);
}
```

I only get:  

    [Standard]

as output. Am I missing someting?

fabianmoronzirfas:
The result of these `app.documentPresets` is what is defined here  

bildschirmfoto 2017-07-14 um 08 38 57

basil.js
trych: (Sorry, for the late response, I am on holiday now 😎 )

No, we cannot list them in the docs, as every user can define his own presets with own names (and probably there are a lot of different presets over different versions). So, I would just leave them out and I just put "A4" and "Letter" in there as examples, so people understand right away, what we are talking about.

basil.js
trych: Yes, that's not the page size presets, that's the document presets that also saves stuff like columns and gutter width etc. Since those are not related to page size, I would just leave them out and focus on page size only. Since I don't know how to retrieve all those, I would just leave them out of the error message.

basil.js
fabianmoronzirfas: But how can I list those page size presets? Can you point me to some code?

trych: I don't think you can. I have looked around for a bit as well, and couldn't find anything. You can just catch an error to find out that a page size preset does not exist, afaik.

fabianmoronzirfas: Okay. At least I found where the ones the user creates are stored.  

    "/Users/$USERNAME/Library/Preferences/Adobe InDesign/Version 12.0/de_DE/Page Sizes"

It would be cool to parse these xml files and give this user a hint what presets he made by himself. I'll investigate a bit further to find them. You can create them by hitting `⌘ ⇧ p`  

bildschirmfoto 2017-07-18 um 07 35 31

fabianmoronzirfas: sorry it is `⌥ ⌘ p`

fabianmoronzirfas: I got some update from the Adobe Dev Slack channel by Dirk Becker. These values are not scriptable. 😢  but made a dump of the relevant parts for us. So we have at least a list of them.
[InDesign_kService_PageSizes.c.txt](https://github.com/basiljs/basil.js/files/1154980/InDesign_kService_PageSizes.c.txt)

trych:
Ok, I implemented the page orientation feature, added tests and added quick examples to the docs.

@fabianmoronzirfas I would leave listing of existing page presets out. I don't think it's very important at all, but on the other hand it would be very complicated to include. Also, there are different presets from version to version. The ones that you linked in the text file are not matching the ones that I have in my CS6 installation at all, for example.

This passes all tests and is ready for review. 🔍 🐞

fabianmoronzirfas:
I like what you implemented but find it a bit obscure that we only can't verify if it works by catching an error.

trych:
I agree that it would be nicer to do it another way, but as long as it is not possible …
Or do you know another way?

I mean the user will not know, so it should not matter for user experience. There are a few `try … catch` statements in basil, so I think it should be fine.

fabianmoronzirfas:
I don't know another way. In the outcome it is irrelevant for the user. try catch just feels like using a crowbar to open a door. 😁

ffd8:
Sounds great, but to be honest, I (+ students) never change the document size in code. It's decided before starting, however when doing such a change, this would be very useful. I also like the idea of `b.PORTRAIT` and `b.LANDSCAPE` as optional inputs (perhaps defaulting with portrait?).

PS - if you don't already have a list of page sizes, perhaps this one can help:
`b.inspect(app.printerPresets[0].paperSizeList);`?

trych:
> Sounds great, but to be honest, I (+ students) never change the document size in code.

That's true, but I think the `b.size()` function is actually really useful, if you want to share your basil scripts and have a certain target size in mind. The person who downloads your script would not know how to set up the document, so it is useful to have the script itself take care of setting the size.

I think the printerPreset's paper size list only lists paper formats specific to a printer and does not reflect the page size presets that are available under the document preferences.

Btw. default is neither `b.PORTRAIT` nor `b.LANDSCAPE` but rather the orientation that the document already has. So if you use `b.size("A3")` on a document that is currently in portrait A4, it will stay portrait.

ffd8:
Aha – that's a good example for using `b.size()`. 
And true, that's the paper sizes for printing.. just thought it might be a useful start.

trych:
@b-g @fabianmoronzirfas @ffd8 

This has been waiting here for a while, but should actually be good to go as well, unless you guys have some objections?

PR #201, opened 2017-07-30
Rewrote b.text() so it won't stop a script anymore if font is missing

### LINKED ISSUE #119: Better handling of missing fonts
If a font is called by `b.textFont()` that does not exist on the current system the entire script stops. However, when teaching/learning basil.js in class this is a situation that happens *all. the. time.*

People write little scripts that use a specific font and if someone else wants to try the scripts then either they have to get the font file from somewhere or they have to comment out all the font lines. This situation happens so much that it becomes really annoying to always comment out the lines (and then to remember to reset it in the end).

I think it would be nicer if the script would still execute, fall back to the default font and instead give a strong warning (maybe even with an alert window?) that font xyz is missing and that the visual result of the script will likely be different than intended.

trych:
As discussed in #119, `b.textFont()` does no longer stop the script if a font is missing, but gives a warning instead.

It looks like this:

```js
b.textFont("myFont", "Bold"); // ### Basil Warning -> b.textFont(), font "myFont  Bold" not installed. Using current font instead.
```

It handles cases where a font *is* installed, but does not have any default font style (Regular/Normal/Roman/Book), like this:

```js
b.textFont("Erbar"); // ### Basil Warning -> b.textFont(), font "Erbar  Regular" not installed. Using current font instead.
```

If it does not find the font, it will use the current font instead, which is either the default font or the most recently applied font (if a font has previously been set successfully).

Btw., although it might not look like it from the code, this *does* work for fonts whose default font style is not called `Regular`. So, if I have for example `Arnhem`, whose default style is called `Normal`, this will apply that default style as expected:

```js
b.textFont("Arnhem");
```
Added tests as well, clears all other tests.

Please review! 🔍 🐛 😎 

This closes #119.

### DIFF
--- a/basil.js
+++ b/basil.js
@@ -4964,7 +4964,7 @@ pub.textFont = function(fontName, fontStyle) {
   }
 
   if(app.fonts.itemByName(fontName).status !== FontStatus.INSTALLED) {
-    warning("b.textFont(), Font \"" + fontName + "\" not installed. Using current font instead.");
+    warning("b.textFont(), font \"" + fontName + "\" not installed. Using current font instead.");
   } else {
     currFont = fontName;
   }
--- a/basil.js
+++ b/basil.js
@@ -4954,12 +4954,21 @@ var isValid = function (item) {
  * @return {String} currFont The name of the current font
  */
 pub.textFont = function(fontName, fontStyle) {
-  if (arguments.length === 1) {
+
+  if (arguments.length === 2) {
+    fontName = fontName + "\t" + fontStyle;
+  } else if (arguments.length === 1) {
+    fontName = fontName + "\tRegular";
+  } else {
+    error("b.textFont(), wrong parameters. Use: fontName, fontStyle. fontStyle is optional.");
+  }
+
+  if(app.fonts.itemByName(fontName).status !== FontStatus.INSTALLED) {
+    warning("b.textFont(), Font \"" + fontName + "\" not installed. Using current font instead.");
+  } else {
     currFont = fontName;
   }
-  if (arguments.length === 2) {
-    currFont = fontName + "\t" + fontStyle;
-  }
+
   return currFont;
 };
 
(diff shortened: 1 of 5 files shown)


fabianmoronzirfas:
This works for me. Can't see any problems. 👎 👌 😃
Just an addition. Would be cool to report to the user which font is being used instead. So the warning could be:  

        warning("b.textFont(), font \"" + fontName + "\" not installed. Using current font " + currFont.replace("\t", " ") + " instead.");

trych:
> Would be cool to report to the user which font is being used instead

Good idea, will prepare another commit, so don't merge yet!

trych:
Ok, the fix itself was easy, however, while fixing it, I realized that the global `currFont` variable is saved as a string (the name of the font), instead of a proper font object. I think this should be fixed: it's not very clean and for example with Fabian's `currFont.replace("\t", " ")` from above it returns `Arnhem Regular` when it should instead return `Arnhem Normal`.

So, I will work some more on this and replace the `currFont` string by a font object instead. Then we could do something like

```js
currFont.fontFamily + " " + currFont.fontStyleName
```
to retrieve the correct name.

So, please, don't merge yet, still working on the commit. :)

trych:
Alright, I fixed this.
`currFont` is now a font object globally. All tests are still passed and also stuff like

```b.typo(textFrame, "appliedFont", "Helvetica");```

still works as expected.

The wording of the warning is now:

```### Basil Warning -> b.textFont(), font "FantasyFont Regular" not installed. Using current font "Arnhem Normal" instead.```

@fabianmoronzirfas If you don't find any errors anymore, you could go ahead and merge.

fabianmoronzirfas:
I'm packing my stuff to go to the Ostsee tomorrow morning. Don't know if I can review it tonight. What about @b-g and @ffd8

b-g:
@trych Many thanks for this! Code and tests look good! Happy to merge it!

trych:
Great thanks! Everybody seems to be a bit in summer break mode at the moment. 😎

fabianmoronzirfas:
> Everybody seems to be a bit in summer break mode at the moment. 😎  

Yes indeed. 🌤  🏊  🏖

Issue #123, opened 2016-11-07
b.SCRIPT_NAME


trych:
Introduce a constant `b.SCRIPT_NAME` that contains the name of the current script as a string.

This could be used for the following things:

1. To give a more precise message in InDesign's Undo-Command (that's where the idea originally came from). So the Menu would not state `Undo: Script` but rather `Undo: helloWorld.jsx` which is nice and useful if several scripts are execute in sequence.

2. For the basil.js log message in the console.

3. For the potentially upcoming [`b.log()`](https://github.com/basiljs/basil.js/issues/50).

4. For users to use it for whatever.

fabianmoronzirfas:
I'd suggest `b.SCRIPTNAME` (without underscore). But I think this needs to be a function of some kind. So `b.scriptname()` ?

    b.SCRIPTNAME = File($.fileName).exists === true ? File($.fileName).name : ($.writeln('error message'));

This gives me the error message. Which is great but if we populate b.scriptname from Basil.js which is included I think we will get "Basil.js" as result. Not the user script.

b-g:
+1 for `b.SCRIPTNAME` (without underscore)!
But should we not get rid of the file extension `(new File($.fileName)).name.slice(0, -4)`, so you can use `b.SCRIPTNAME` to e.g. name things without fiddling around ...

fabianmoronzirfas:
> But should we not get rid of the file extension   

Good point.  

BTW: You don't need to use the `new` to get a `File` that already exists.

fabianmoronzirfas:
Is the slice(0, -4) bullet proofed against `.js` or even files without extension? I think we should use some kind of regex for this  like this example http://stackoverflow.com/a/4250408/1770432

trych:
Yes, leaving out the extension is a good idea and using a regex for handling both `.jsx` and `.js` also.

Btw., there is also `app.activeScript` for "The current script running from the Scripts panel." Is this what we need or `$.fileName`?

fabianmoronzirfas:
Well we should try it out. I always us the `$.fileName` but only because it works on all applications. (e.g. AE, AI etc)

b-g:
to summarize the conversation:
 
decision: b.SCRIPTNAME 
decision: use a regex to be on the safe side
decision: for `$.fileName` as `app.activeScript` is not working is you execute a script from outside e.g. via Sublime.

trych:
This just got merged. Closing.

PR #202, opened 2017-08-13
b.SCRIPTNAME

trych:
As discussed in #123, this adds a constant `b.SCRIPTNAME` that holds the name of the script. This will change the welcome text from the current

```Running in basil.js 1.1.0 ...```
to
```Running ellipses.jsx using basil.js 1.1.0 ...```

Also, it will add the name of your script to the undo menu:

bildschirmfoto 2017-08-13 um 15 38 04

As you can see, I left the file extension in there, because I was trying it out a bit and it felt weird if it's missing:

bildschirmfoto 2017-08-13 um 15 32 45
bildschirmfoto 2017-08-13 um 15 35 00

I think it might not be obvious right away that his refers to a script named `ellipses`. As opposed to this:

bildschirmfoto 2017-08-13 um 15 37 57
bildschirmfoto 2017-08-13 um 15 38 04

I think here you understand much better that this refers to the script. Also, we can always use basil's own `b.split()` method to retrieve the script's name without extension:

```var scriptNameWithoutExtension = b.split(b.SCRIPTNAME, ".")[0]```

Let me know, if you disagree. 😃

If you look at the code, you can see that neither `$.filename` nor `app.activeScript` was used to retrieve the name, as they both did not return the script name, so I had to evaluate the stack trace, which is the only way I found to return the name reliably.

Please review. 🔍 🐛

### DIFF
--- a/basil.js
+++ b/basil.js
@@ -337,16 +337,15 @@ pub.AFTER = LocationOptions.AFTER;
 pub.LOREM = "Lorem ipsum is dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.";
 
 /**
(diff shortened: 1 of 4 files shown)


b-g:
@trych Many thanks for this! Really great hack/find/solution. I ran quite some time into the inconsistency of $.filename and app.activeScript. Yay! + merged!

trych:
Thanks!

Issue #203, opened 2017-08-16
Checking for `currPage` and `currDoc` fails in certain situaions.


trych:
As I'm trying to implement `b.revert()`(#133) , I ran into a bug.

Background: If you have saved the current page in a variable `currPage` and *then* revert the doc, the variable will still hold a page object, but at the same time it returns that this page object equals null.

You can test this small script to check for yourself:

```js
// @target InDesign

var doc = app.activeDocument;
var currPage = doc.layoutWindows[0].activePage;

// add something, so the document can be reverted
currPage.ovals.add();

doc.revert();

$.writeln("currPage: " + currPage);  // [objectPage]
$.writeln("currPage instanceof Page: " + (currPage instanceof Page));  // true
$.writeln("currPage === null: " + (currPage === null));  // true (!)
```

This makes the checks for [currPage](https://github.com/basiljs/basil.js/blob/f34b8a3722dd635dc0dd7cc45edfc595865e1f58/basil.js#L1039) and the [currDoc](https://github.com/basiljs/basil.js/blob/f34b8a3722dd635dc0dd7cc45edfc595865e1f58/basil.js#L912) fail, as they do not enter the `if(!currPage)` if clause and after that mess up pretty badly (as basil.js) continues while thinking it holds a valid page object.

I therefore propose that the checks will be changed to

```
if(currPage === null)
```

or, to be on the safe side

```
if(currPage === null || !currPage)
```
(although I don't think that it ever receives anything else apart from `null` or `[object Page]`.

This post serves more as a reference for the person that will review my `b.revert()` PR once it's done, to explain why I changed these checks. 😉

fabianmoronzirfas:
Okay. So we can close this one. The changes are in #204

Issue #133, opened 2016-11-14
Universal way to reset documents: b.revert()


trych:
I think I now found a good solution to a problem that has been bugging me since July and that I have already described in #78.

While developing a script it is of great help to put `b.clear(b.doc());` at the beginning of the script, so the script starts with a cleared document each round. However, if the script that is developed contains steps like creating new pages, layers, colors etc. this method fails, as these are not cleared by `b.clear()`. So in this case you either have to do a manual undo after each execution or you have to close the document and open a new one, which you could also do with `b.close(false);`.

`b.close(false);` however would not work on documents that I have saved before executing (for example when I need to use basil's data folder). So I have to manually undo the script or revert to the last save.

Therefore I propose a function `b.revert()` that does exactly that, to have one universal command to always reset the document to it's initial state (as InDesign's `Revert` command does). So, if I have my document saved, it returns to the last saved state. If I have my document not saved yet *it would close the unsaved document without saving and open a new empty document*, kind of "reverting to the initial empty document".

That way I could always just put a `b.revert()` command at the top of my script during development and not have to worry if it is a script with a saved doc, a script that adds layers or pages or a script that is not saved yet (and therefore always starts from a new empty document).

Thoughts? 😊

fabianmoronzirfas:
Sounds great. Way better then using `b.clear` with additional parameters

trych:
Bonus: it would also be much faster than `b.clear(b.doc())` and you could very nicely and quickly create 100s of custom PDFs from scratch like so:

```js
for(var i = 0; i < 100; i += 1) {
  b.revert();
  // do some stuff, maybe even add random colors or layers or stuff
  b.savePDF(i + ".pdf", false);
}
```

b-g:
Sound great +1!

But should it not be `b.revert(b.doc())`? That way you have at least a hint what is reverted ...

trych:
> But should it not be b.revert(b.doc())?

I think as long as there are not any options of what is going to be reverted there should be no argument. As a doc is the only thing that *can* be reverted, I would leave it without argument. This would be also consistent with functions like `b.close()`, `b.size()` etc.

b-g:
@trych OK + convinced! Just `b.revert()` is better in terms of consistency.

fabianmoronzirfas:
Should `b.revert()` return the current doc?  
e.g. When using revert on an unsaved document. Wouldn't it be nice to directly get the new doc from revert?

trych:
[...]

PR #204, opened 2017-08-16
Adding b.revert()

trych:
This adds `b.revert()` as proposed and discussed in #133.

As a reminder: This is useful during the development stage of a script, if you don't want to reset a document manually each time before you run the script (`b.clear(b.doc())` would not work in many cases, as it does not clear new layers and colors etc.)

So while developing your script, it would simply look like this:

```js
// @includepath "~/Documents/;%USERPROFILE%Documents";
// @include "basiljs/basil.js";

function draw() {

  b.revert();
  
  // all code that you are developing here

}

b.go();
```
And this would work both for saved and unsaved documents.

I did some testing and figured out that it is actually much quicker to close the document without saving and re-opening it than use InDesign's actual `doc.revert()` method, so I went with that. Tested this both with empty documents and with larger documents.

This modifies some conditional checks as well, which I explain in #203.

Added tests for it, it passes all other tests as well.

Please review. 🔍 🐛 😎

### DIFF
--- a/basil.js
+++ b/basil.js
@@ -1457,29 +1457,6 @@ pub.clear = function(container) {
 };
 
 /**
- * @description Reverts the document to its last saved state. If the current document is not saved yet, this function will close the document without saving it and reopen a fresh document so as to "revert" the unsaved document. This function is helpful during development stage to start from a new or default document each time the script is run.
- *
- * @cat Document
- * @method revert
- * @return {Document} The reverted document.
- */
-pub.revert = function() {
-
-  if(currDoc.saved && currDoc.modified) {
-    var currFile = currDoc.fullName;
-    currDoc.close(SaveOptions.NO);
-    app.open(File(currFile));
-    resetCurrDoc();
-  } else if(!currDoc.saved) {
-    currDoc.close(SaveOptions.NO);
-    resetCurrDoc();
-    currentDoc();
-  }
-
-  return currDoc;
-};
-
-/**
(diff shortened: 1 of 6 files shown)


fabianmoronzirfas:
Cool thing. To things I noticed when trying it out

`b.revert()`  

Creates a new doc when no document is open.

Expected outcome:  

Doing nothing. Or warning the User that there is no document to work on.

When more then one document (unsaved) is open, the document is closed and not recreated. See the recording.   
revert


Expected outcome:

A new blank doc shoud be created.

trych:
Hi Fabian, thanks for giving it a try.

> b.revert()
> Creates a new doc when no document is open.

Actually basil.js *always* creates a new document, when you run it, even if you run the empty basil.js template. So, at the moment when the script reaches `b.revert()` there is already a document open (and yes, this gets reverted then).

So, I don't think it's a problem (also since this does usually only happen, if you run the script for the first time during developing stage. After developing stage it makes more sense to take `b.revert()` out (except you need it to reset the document between exports etc.).

> When more then one document (unsaved) is open, the document is closed and not recreated. See the recording.

Ok, this is not expected indeed, I will investigate, thanks!

trych:
> When more then one document (unsaved) is open, the document is closed and not recreated. See the recording.

Ok, this remaining issue should be fixed now. If there are no other issues, this could be merged.

b-g:
@trych Many thanks + looks good! Merged!

Issue #145, opened 2016-11-30
Convert graphic to text and visa-versa


ffd8:
While testing the needs for a b.textMode(b.CENTER) « so one can draw text frames from center out... I realized that we don't have an easy way to place text into a normal rect = [converting a graphic to text frame](https://forums.adobe.com/thread/611553). 

Therefore I'd suggest a basic contentType switcher. Routes I could imagine going down are:

Processing + beginner style:
`b.toText(obj);`
`b.toGraphic(obj);`

`b.contentType(obj, b.CONSTANT);`
`b.TEXT_TYPE`, `b.GRAPHIC_TYPE`, `b.UNASSIGNED`

In the background, they do the same thing, ie:

`obj.contentType = ContentType.TEXT_TYPE; `
`obj = obj.getElements()[0];`

I've recalled the past two weeks how 1st timer struggle with knowing the right wording/capitalization for constants.. but they do let us pack things into single function. Started off for the `b.toText()` style.. but slowly leaning towards `b.contentType()` – since one has to read the reference/docu anyways to know about the function.

Thoughts?

b-g:
@ffd8 Interesting! Never had that problem.

Q: Are the two type castings `Graphic > Text` and `Textframe > Graphic` the only cases?

Also I find the indesign wording confusing as my first association with "TEXT" was not a Textframe but rather the text inside a textframe.

trych:
Could this maybe be implemented into `b.text()` and `b.image()` directly?
What we can do at the moment is this

```js
var myEllipse = b.ellipse(20, 20, 80, 80);
var myImage = b.image("img.jpg", ellipse);
```
So we might as well also allow stuff like

```js
var myTextFrame = b.text(b.LOREM, 20, 20, 80, 80);
var myImage = b.image("img.jpg", myTextFrame); // --> turns myTextFrame into graphic frame
```

and

```js
var myEllipse = b.ellipse(20, 20, 80, 80);
var myImage = b.text(b.LOREM, myEllipse); // --> places text into the ellipse
```

On your other point

> While testing the needs for a `b.textMode(b.CENTER)`

thought about this as well recently. I actually always wondered, why textFrames don't use the currently set `rectMode` by default. I would find this the most intuitive solution (yes, I know a `textFrame` is not a `rect`in InDesign terms, but it is in fact a rectangle in the end. ;) )

fabianmoronzirfas:
what about having a new function called something like 

    b.place(filepath | String , frame);

[...]

PR #208, opened 2017-10-04
Shape Mode fixes

trych:
This PR fixes the following:

- Adds the `b.RADIUS` option for `b.rectMode()` (for some reason this was missing, although it was mentioned in the docs?).
- Makes `b.rectMode()` work for text frames as well (as we agreed on [here](https://github.com/basiljs/basil.js/issues/145)). This makes working with both text frames and rectangles much more intuitive, as they are drawn in the same way.
- Updates, improves and fixes docs for `b.rectMode()`, `b.ellipseMode()`, `b.text()`.

Please review! 🔍 🐛


trych:
Ok, resolved the minor `changelog.txt` merge conflict. Merging this finally.

Issue #136, opened 2016-11-18
b.pageCount() as setter


trych:
Small improvement: `b.pageCount()` could also act as setter (it's a getter only at the moment). Most getters work as setters in basil, would be nice, if the same worked for `b.pageCount()`.

```js
b.pageCount(20); // -> changes document to have 20 pages
```

fabianmoronzirfas:
What happens if I set a number lower then my current page count? Will it delete pages?

fabianmoronzirfas:
How does it handle master spreads?

trych:
I think it should just adopt the behavior of `b.doc().documentPreferences.pagesPerDocument = n`.

> What happens if I set a number lower then my current page count? Will it delete pages?

Yes, if I add a line `b.pageCount(5)` at a point in a script where I know I have a 20 page document, I should expect to have pages 6 - 20 deleted.

> How does it handle master spreads?

Just as `b.doc().documentPreferences.pagesPerDocument = n` would as well. I believe the behavior is to have the master page of the currently active page applied to the new pages.

I just think it is a simple addition that would be mostly used in document setup. As beginners don't know about (the very non-basil) `b.doc().documentPreferences.pagesPerDocument` the standard way of getting a 20 page document seems to be:

```js
for(var i = 1; i < 20; i++) {
    b.addPage();
}
```

So as `b.pageCount()` already exists anyway, it would be nice to be able to just do `b.pageCount(20)` instead.

ffd8:
Seems a tad dangerous that it can just remove pages by giving the potentially wrong number (counting accidents happen.. especially with beginners when 0 usually equals the new 1). 

I wonder if such functionality should be merged to the `b.addPage()` and `b.removePage()`? Unfortunately they were built with the single page in mind.. rather than `b.addPages(#, where)`. There are a few functions that already depend on the singular vs plural wording `b.selection()` vs `b.selections()`.. so that could be an option?

trych:
> Seems a tad dangerous that it can just remove pages by giving the potentially wrong number

Hm, I cannot really imagine an actually dangerous real life scenario here, it's not like people run this command on their production files for testing and then save those right afterwards.

If however during script development they figure out that more pages are deleted than intended, then they can simply correct this error until they got it right and then when the script is behaving as intended use it for "production".

I'm pretty sure 99% of the time the feature would be used during "document setup stage" to just give your initial blank document the number of pages you want it to have.

[...]

PR #209, opened 2017-10-17
pageCount can be used as setter now

trych:
This is a quick and easy one, allowing `b.pageCount()` to be used as a setter, as we discussed in #136.

Please review. 🐛 🔍

### DIFF
--- a/basil.js
+++ b/basil.js
@@ -1844,14 +1844,25 @@ pub.previousPage = function () {
 
 
 /**
- * The number of all pages in the current document.
+ * Returns the number of all pages in the current document. If a number is given as an argument,
+ * it will set the document's page count to the given number by either adding pages or removing
+ * pages until the number is reached. If pages are added, the master page of document's last
+ * page will be applied to the new pages.
  *
  * @cat Document
  * @subcat Page
  * @method pageCount
+ * @param  {Number} [pageCount] New page count of the document (integer between 1 and 9999).
  * @return {Number} The amount of pages.
  */
-pub.pageCount = function() {
+pub.pageCount = function(pageCount) {
+  if(arguments.length) {
+    if(pageCount.constructor.name === "Number" && pageCount % 1 === 0 && pageCount > 0 && pageCount < 10000) {
+      currentDoc().documentPreferences.pagesPerDocument = pageCount;
+    } else {
+      error("b.pageCount(), wrong arguments! Use an integer between 1 and 9999 to set page count.");
+    }
+  }
   return currentDoc().pages.count();
 };
 
--- a/basil.js
+++ b/basil.js
@@ -1846,7 +1846,7 @@ pub.previousPage = function () {
 /**
  * Returns the number of all pages in the current document. If a number is given as an argument,
  * it will set the document's page count to the given number by either adding pages or removing
- * pages until the number is reached. If pages are added, the master page of document's last
+ * pages until the number is reached. If pages are added, the master page of the document's last
  * page will be applied to the new pages.
  *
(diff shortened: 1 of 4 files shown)


ludzeller:
Cool stuff!

basil.js
```diff
@@ -1844,14 +1844,25 @@ pub.previousPage = function () {
...
  * @cat Document
  * @subcat Page
  * @method pageCount
+ * @param  {Number} [pageCount] New page count of the document (integer between 1 and 9999).
  * @return {Number} The amount of pages.
  */
-pub.pageCount = function() {
+pub.pageCount = function(pageCount) {
+  if(arguments.length) {
+    if(pageCount.constructor.name === "Number" && pageCount % 1 === 0 && pageCount > 0 && pageCount < 10000) {
```
fabianmoronzirfas: Looks great. I just needed to look up what  `pageCount % 1 === 0` does. I guess this is a thing we should not forget to do. Write comments into our code to allow future maintainers to not have to google for [sto questions](https://stackoverflow.com/questions/3885817/how-do-i-check-that-a-number-is-float-or-integer).  😄

fabianmoronzirfas: Also wouldn't it be more readable if it was a bit more like this:  

```js
if(arguments.length > 0) {
	if(pageCount.constructor.name === "Number"){
		if(pageCount % 1 === 0){
			if (pageCount > 0 && pageCount < 10000) {
				
			}
		}
	}
}
```

I think this is mostly a question of style. I always dislike cramming to many conditions into one `()`

trych: Damn, how did you find my secret source. 😆

trych: Hm, I personally find 4 nested if statements to be less readable than 4 chained ones. Yes, it's not ideal to have to many conditions in one `()`, but nesting 4 levels deep is not ideal either.

We could maybe add an internal helper function `isInteger()`. Then it would be a bit more easy to read and we don't need to put a comment somewhere to make it even more messy.

Then it would read
```js
if(isInteger(pageCount) && pageCount > 0 && pageCount < 10000) {

}
```

fabianmoronzirfas: Yeah. Maybe that’s a way. 4 nested conditions is to much. That’s right. I just think it should be more expressive to get the gist.

trych: Sure thing. I'll add `isInteger()`. If we have it anyways, should we also expose it to the user? We have `isNumber()`, `isString()` and `isArray()` already, so I guess there would be no harm in adding another one. What do you think?

fabianmoronzirfas: Sure. Why not.

trych:
Ok, added an `b.isInteger()` to shorten the if condition (and as an extra goodie for the end user 😉 ).

Please review. 😎

fabianmoronzirfas:
Looks good to me. Any other objections? @basiljs

trych:
Who actually is @basiljs ? 😄

b-g:
ATbasiljs = I guess this is all of use! + Great work, no comments from my side! IMO just press merge any time! thanks!

trych:
Done.

fabianmoronzirfas:
> Who actually is @basiljs ?  

Should be a way to notify the whole team. Everyone should get a message if you use this.

trych:
> Should be a way to notify the whole team.

Really? I didn't know that. I always thought ATbasiljs was a user. Cool!

fabianmoronzirfas:
> Really? I didn't know that. I always thought ATbasiljs was a user. Cool!  

https://github.com/blog/1121-introducing-team-mentions

PR #210, opened 2017-10-18
Fixing b.group() and b.ungroup()

trych:
I just came across these two functions and the docs were seriously messed up/mixed together (In the auto generation of docs (two `b.group()` methods were created due to a wrong `@method` parameter and lots of other stuff).

So, here are the improvements:

- Fixed and improved the docs
- Made error messages more understandable
- Added additional error checking if strings of non-existing groups are given
- Cleaned up the methods

This does not really change anything in functionality (except the doc generation), so a quick review should do.

### DIFF
--- a/src/includes/environment.js
+++ b/src/includes/environment.js
@@ -461,32 +461,36 @@ pub.layer = function(layer) {
 
 
 /**
- *  Returns the Group instance and sets it if argument Group is given.
+ *  Groups items to a new group. Returns the resulting group instance. If a string is given as the only
+ *  argument, the group by the given name will be returned.
  *
  *  @cat Document
  *  @subCat Page
- *  @method Group
- *  @param {Array} [pItem] The PageItems array (must be at least 2) or name of Group name instance.
- *  @param {String} name The name of the Group, only when creating a Group from Page Item(s).
- *  @return {Group} The current Group instance.
+ *  @method group
+ *  @param {Array} pItems An array of page items (must contain at least two items) or name of group instance.
+ *  @param {String} [name] The name of the group, only when creating a group from page items.
+ *  @return {Group} The group instance.
  */
-pub.group = function (pItem, name) {
-  checkNull(pItem);
-  var group = null;
-  if(pItem instanceof Array) {
-    if(pItem.length < 2) {
-      error("There must be at least two PageItems passed to b.group().");
+pub.group = function (pItems, name) {
+  checkNull(pItems);
+  var group;
+  if(pItems instanceof Array) {
+    if(pItems.length < 2) {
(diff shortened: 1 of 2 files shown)


src/includes/environment.js
```diff
@@ -461,61 +461,68 @@ pub.layer = function(layer) {
...
  */
-pub.group = function (pItem, name) {
-  checkNull(pItem);
-  var group = null;
-  if(pItem instanceof Array) {
-    if(pItem.length < 2) {
-      error("There must be at least two PageItems passed to b.group().");
+pub.group = function (pItems, name) {
+  checkNull(pItems);
+  var group;
```
fabianmoronzirfas: Why not setting it `null`? I often do that to not have an undefined variable

trych: Hm, I only set variables to `null`, if I need them to be `null` for some reason. Here I don't need them to be anything for the moment, I just need to initialize them, so I do it in the shortest, most concise way possible.

Assigning `null` would feel a bit like.

```js
var group = "temp";
```

As in "I just assign some value, although I don't need it anyways".

Also, without assigning `null`, I find it is much cleaner and easier to initialize several variables:

```js
var foo, bar, baz;
```

I just checked the basil.js source and there are many places, where `var foo;` is used to initialize.

What do you think?

fabianmoronzirfas: I like my variables to be `null`. :-) So I can always check if they are null or something else. I found `undefined` to be sometimes not reliable for checking. Don't bother. Just a question.  

It could also be:  

    var x = y = z = null;

fabianmoronzirfas: Assigning null is for me different then giving a string a content. On a string I would go for 

    var thing = ''; // or null

trych:
```js
var x = y = z = null;
```
Yes, I don't find this as clean though. No matter, if `undefined` is reliable for checking or not, in this case I know that I *don't* need to check it, that's why I think, I don't need to assign anything, not even `null`. And you're right, the string comparison does not really work.

I guess it just comes down to style. Googling also revealed that some do it like this, others like that.

> Don't bother. Just a question.

Cool, so I will just keep it like this now, okay?

fabianmoronzirfas: > Cool, so I will just keep it like this now, okay?
Yes. If not we should change things like this throughout basil.

trych: Ok, good to merge then, right?

fabianmoronzirfas: For me. Yes. 👍

trych:
This should be ready to go, unless somebody has some objections?

b-g:
Thanks! Fine for me. (Plz merge)

Issue #213, opened 2017-11-08
b.startsWith and b.endsWith give cryptic error on passing objects


trych:
When `b.startsWith()` or `b.endsWith()` are used with a (word, paragraph, etc. ...) object instead of a string it gives the cryptic error message:

```js
ReferenceError: str.indexOf is not a function
```
There should be a more meaningful error message, as this error happens all the time.

fabianmoronzirfas:
Yes it should. I think it is the overall try catch that produces that.

trych:
Yes, we just need to check the inputs for actually being strings.
Btw., I also used this as an in-class demo how to write issues on GitHub. ;)

fabianmoronzirfas:
> Btw., I also used this as an in-class demo how to write issues on GitHub. ;)
double + good!

trych:
Fixed in #215 .

PR #215, opened 2017-11-13
Improved error handling for b.startsWith() and b.endsWith()

trych:
A quick fix to add some parameter checks for `b.startsWith()` and `b.endsWith()` as described in #213. Quick review anyone.? :)


trych:
@basiljs Could I go ahead and merge this?
It surely would impress the crap out of my students if the issue that we found and reported in class last week would be fixed in the following week. 😉

b-g:
😂👍! Plz!

Issue #214, opened 2017-11-08
Rounding option for rectangle corners


trych:
This came up in class as well today.

In processing and in p5.js the `rect()` commands already have the option to create round corners. basil.js should also include this, I think.

Could be used for stuff like [this](https://i1.wp.com/daks2k3a4ib2z.cloudfront.net/55459cebb2d38dc972871f70/554ad41cabf193924ca3e03d_01-Sufjan-Stevens-Carrie-%26-Lowell-20150605.jpg?resize=755%2C1111&ssl=1).

fabianmoronzirfas:
```
Syntax	
b.rect(a, b, c, d)
b.rect(a, b, c, d, r)
b.rect(a, b, c, d, tl, tr, br, bl)
Parameters	
a	float: x-coordinate of the rectangle by default
b	float: y-coordinate of the rectangle by default
c	float: width of the rectangle by default
d	float: height of the rectangle by default
r	float: radii for all four corners
tl	float: radius for top-left corner
tr	float: radius for top-right corner
br	float: radius for bottom-right corner
bl	float: radius for bottom-left corner
```
[source](https://processing.org/reference/rect_.html)

trych:
Implemented via #217 and successfully used in my class right now. 😜

Closing this.

PR #217, opened 2017-11-13
b.rect() with rounded corners

trych:
This adds the option to draw your rects with rounded corners.

This
```js
  b.noFill();

  b.rect(30, 20, 80, 55);

  b.rect(30, 90, 80, 55, 20);

  b.rect(30, 160, 80, 55, 20, 15, 10, 5);
```
turns into this
bildschirmfoto 2017-11-13 um 18 04 17

So as Fabian summed up nicely [here](https://github.com/basiljs/basil.js/issues/214), syntax is:
```
Syntax	
b.rect(a, b, c, d)
b.rect(a, b, c, d, r)
b.rect(a, b, c, d, tl, tr, br, bl)
Parameters	
a	float: x-coordinate of the rectangle by default
b	float: y-coordinate of the rectangle by default
c	float: width of the rectangle by default
d	float: height of the rectangle by default
r	float: radii for all four corners
tl	float: radius for top-left corner
tr	float: radius for top-right corner
br	float: radius for bottom-right corner
bl	float: radius for bottom-left corner
```

Please review. 🔍 🐛 

This is another one we discussed in class last week, so if we could get this merged before my class on Wednesday, again, I could show my students the perks of open source development. 😄

### DIFF
--- a/src/includes/shape.js
+++ b/src/includes/shape.js
@@ -420,14 +420,14 @@ pub.rect = function(x, y, w, h, tl, tr, br, bl) {
   }
 
   if(arguments.length > 4) {
+    for(var i = 4; i < arguments.length;i++){
+      if(arguments[i] < 0 ){
+        error("b.rect(), needs positive values as arguments for the rounded corners.");
+        return;
+      }
+    }
     newRect.topLeftCornerOption = newRect.topRightCornerOption = newRect.bottomRightCornerOption = newRect.bottomLeftCornerOption = CornerOptions.ROUNDED_CORNER;
     if(arguments.length === 8) {
-      for(var i = 4; i < arguments.length;i++){
-        if(arguments[i] < 0 ){
-          error("b.rect(), needs positive values as arguments for the rounded corners.");
-          return;
-        }
-      }
       newRect.topLeftCornerRadius = tl;
       newRect.topRightCornerRadius = tr;
       newRect.bottomRightCornerRadius = br;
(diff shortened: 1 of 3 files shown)


fabianmoronzirfas:
The only thing I can find is that providing negative values returns an basic indesign error.  

bildschirmfoto 2017-11-13 um 19 30 12

Would be nice to have a better message.

src/includes/shape.js
```diff
@@ -422,6 +422,12 @@ pub.rect = function(x, y, w, h, tl, tr, br, bl) {
   if(arguments.length > 4) {
```
trych: How about instead placing it after *this* line (so if using one radius, it will also be checked)? Also, what about this syntax:

```js
if( pub.min(Array.prototype.slice.call(arguments, 4)) < 0 ) {
  error("b.rect(), needs positive values as arguments for the rounded corners.");
}
```

fabianmoronzirfas:
Moved it up. I have no feelings about how to write it. What's the difference?

src/includes/shape.js
fabianmoronzirfas: Moved it up. I have no feelings about how to write it. What's the difference? Is it faster?

trych: No, I thought maybe mine is a bit shorter. Yours is more legible though, so just keep it. The only thing is, that I would get rid of the `return` statement. We generally don't use return after throwing errors.

fabianmoronzirfas: Old habit. Get of my lawn you damn kids!

trych:
Ok, just bundled it now, so it is good to go. Will merge.

PR #224, opened 2017-11-27
b.arrange()

### LINKED ISSUE #109: b.arrange()
I think it would be useful to have the possibility to arrange objects in z-order after creating them.

I am not sure, what the most elegant way would be to implement this. We _could_ use the constants `b.AT_END`, `b.AT_BEGINNING`, `b.AFTER` and `b.BEFORE` to describe `item.sendToBack()`, `item.bringToFront()`, `item.sendBackward()`, `item.sendForward()`.

Or we could use these four constants only if there is a reference object to move the object behind or in front of and then introduce two additional constants `b.BACKWARD`, `b.FORWARD`.

The method could work like this:

`b.arrange(objectToMove, locationOrDirection, [referenceObject]);`

trych:
This adds `b.arrange()` to allow to sort page items and layers in z-space.
I just quote the docs to explain functionality:

Arranges a page item or a layer before or behind other page items or layers. If using the constants `b.FORWARD` or <code>b.BACKWARD</code> the object is sent forward or back one step. The constants <code>b.FRONT</code> or <code>b.BACK</code> send the object to the very front or very back. Using <code>b.FRONT</code> or <code>b.BACK</code> together with the optional reference object, sends the object in front or behind this reference object.

So you have commands like
```js
b.arrange(myPageItem, b.FRONT);
b.arrange(myPageItem, b.BACK, myReferenceItem);
b.arrange(myLayer, b.BACKWARD);
b.arrange(myLayer, b.FRONT, myReferenceLayer);

// etc.
```
I set up an [example script](https://github.com/basiljs/examples/blob/arrange/environment/arrange.jsx) to demonstrate all arrange options for pageItems:

arrange

Would be great if one of you guys could give this a quick review. 🐛 🔍  If all seems fine, it can be merged right away. Thanks! 😎 

Edit: This closes #109. (Holy cow, that's a lot of issues ago now)


ffd8:
great addition! the example with 'myReferenceItem' doesn't seem so natural, but nevertheless a very useful addition to have to the function. thanks!

trych:
Thanks!

> the usage with 'myReferenceItem' doesn't seem so natural

Yes, this basically imitates the InDesign scripting way of doing it. I guess most of the times the method will be used to just send items all the way to the back.

PR #228, opened 2017-12-01
Simplified isString

trych:
Made isString easier to read and (ever so slightly) faster.

### DIFF
--- a/basil.js
+++ b/basil.js
@@ -3454,7 +3454,7 @@ var isInteger = pub.isInteger = function(num) {
  * @return {Boolean} returns true if this is the case
  */
 var isString = pub.isString = function(str) {
-  return Object.prototype.toString.call(str) === "[object String]";
+  return str.constructor.name === "String";
 };
 
 /**
--- a/src/includes/data.js
+++ b/src/includes/data.js
@@ -713,7 +713,7 @@ var isInteger = pub.isInteger = function(num) {
  * @return {Boolean} returns true if this is the case
  */
 var isString = pub.isString = function(str) {
-  return Object.prototype.toString.call(str) === "[object String]";
+  return str.constructor.name === "String";
 };
 
 /**


fabianmoronzirfas:
Why not:   

     typeof str === "string"

trych:
Well, because I did not think of that. 😜
Just gave it a test, your version is _a lot_ faster. Could you prepare a PR?

fabianmoronzirfas:
hm 

```js
function isstring(str){
    return typeof str === "string" || typeof str === "String"
    }

var foo = new String("String");
var bah = "string";
$.writeln("res foo = " + isstring(foo));
$.writeln("res bah = " + isstring(bah));
```

Gives me 

    res foo = false
    res bah = true

fabianmoronzirfas:
This does it:  
```
function isstring(str){
    return typeof str === "string" || str instanceof String
    }

var foo = new String("String");
var bah = "string";

$.writeln("res foo = " + isstring(foo));
$.writeln("res bah = " + isstring(bah));
```

    res foo = true
    res bah = true

fabianmoronzirfas:
your version gives me both right.

```js
function isstring(str){
    return str.constructor.name === "String";
    }

var foo = new String("String");
var bah = "string";

$.writeln("res foo = " + isstring(foo));
$.writeln("res bah = " + isstring(bah));
```
output:  

```
res foo = true
res bah = true
```

trych:
Well, that's exactly why I picked constructor name. 😉

Haha, no not really. Ok, so we'll leave it at constructor name then, I guess.

fabianmoronzirfas:
Yes it think so. If somebody needs speed he still can implement it or come to us.

trych:
Argh, I just realized that my version does not work with `null` or `undefined` as an input. Back to the original version. 😕

fabianmoronzirfas:
The Version befördert did?

fabianmoronzirfas:
I meant before

trych:
The original worked with all inputs, including `null`, my simplification did not work with all inputs anymore, hence the revert to the original version.

Issue #222, opened 2017-11-27
Getting matrix functionality to work


trych:
## The issue
(I try to describe my understandig of the issue we currently have, please someone edit it to add some missing stuff or to correct me)

Currently we have transformation methods such as `b.translate()`, `b.rotate()` and `b.scale()`. These are – in theory – supposed to mimic the behavior of Processing's equivalents `translate()`, `rotate()` and `scale()`.

In Processing all these methods relate to the origin of the scene (by default [0, 0], _top_ left corner), and they get applied to the shapes that will be added after calling the methods. So, when I do something like

```js
rotate( radians(45) );
rect(50, 0, 20, 20);
```
the rect is not rotated by `45°` around _itself_, but around the origin.

Here's where basil is different at the moment. If I do

```js
b.rotate( b.radians(45) );
b.rect(50, 0, 20, 20);
```

the rect is created at position `[0, 50]` and then rotated by 45 around _itself_ (or rather around its own origin, which is whatever InDesign is currently set to, most probably the default top left corner, but could be any other corner or the center).

`b.scale()` has the same problem that it only relates to the shape's own origin, not to the scene origin. `b.translate()` seems to behave as intended (?), so if I do: 

```js
b.translate(20, 30);
b.rect(0, 0, 50, 50);
```
the rect _does_ get created at `[20, 30]`.

Additionally there are a bunch of methods `b.Matrix2D`, `b.Matrix2D.rotate()` that are supposedly transform the Adobe matrix, but currently nobody really seems to know how to use them or if they ever worked (please correct me, if I am wrong).

## Proposed Solution
The proposal would be to make all three transformations method act like they do in Processing, ideally utilizing Adobe's matrix system and to figure out if (and why) we need separate `b.Matrix2D` methods.

(For transformations around the shape's own origins there will be a new `b.transform(shape, "transformationName", value)` method (that is already taken care of)).

## Resources
[Coordinate Spaces & Transformations in InDesign — Chap.1-3](http://www.indiscripts.com/post/2014/03/coordinate-spaces-and-transformations-1)

[Coordinate Spaces & Transformations in InDesign — Chap.4](http://www.indiscripts.com/post/2016/06/coordinate-spaces-and-transformations-4)

Most certainly these documents give the best overview over the topic, as (quote Marc Autret) _"the fault [of being an obscure topic] mainly lies with Adobe documentation, especially the Scripting DOM reference, which does not clearly explain the topic and some of its essential keys"_.

[...]

PR #236, opened 2017-12-16
Solution for transformation issue #222

bohnacker:
This commit solves the problem, that rotate and scale behave
differently in InDesign and Processing. At least I hope so.


trych:
Oooo, christmas is coming (1 week) early this year. @bohnacker thank you so much for looking into this, really appreciate that you help out with this stuff, that is over our (or at least my) heads.

I will give this some good testing later and report back, if this works for me.

trych:
Ok, I had a closer look at this and it seems to work really, really nicely. Tested a lot of stuff and I am really impressed that with so few lines changed, suddenly everything seems to work!

As there was some stuff missing (matrix fixes for `b.text()`, `b.image()` and `b.duplicate()` as well as the bundling of the basil.js file), I moved your changes to a new [transformation branch](https://github.com/basiljs/basil.js/tree/transformation) to be able to work on all remaining changes.

Before this gets merged, I would actually like to implement the new `b.transform()` in the same branch, because the merge will break pretty much _all_ scripts that ever used `b.transform()` and `b.scale()`. By implementing `b.transform()` users will have a method at hand that works like the old "broken" transforms.

@basiljs What do you guys think? Sounds like a plan?

Thanks again @bohnacker for this great contribution!

fabianmoronzirfas:
applause  

Thank you @bohnacker great work. @trych I think we can wait or merge this right away. It's up to you. I think your students might be the only ones using the develop branch for their work, so if you think that is a better way 👍

b-g:
@bohnacker Yay! Super mega cool! 🙏👏🎄

trych:
Closing this, so the remaining changes can be put in the transformations branch.

Issue #241, opened 2018-01-05
Implementing "b-less"


trych:
Hey everybody!

After the [restructuring issues](https://github.com/basiljs/basil.js/issues/239) I had last week @fabianmoronzirfas and me had a quick Skype call on how to implement b-less.

I have given this some thought over the last few days and I think we have come up with a good solution. This issue serves as a roadmap for ourselves and to inform the rest about the plan.

The plan:

- [x] get rid of `b.` (duh!)
- [x] have all the basil functions live in global space
  - [x] get rid of the ~~`pub` and the ~~`glob` object
  - [x] ~~rewrite all functions from `pub.rect = function(args)` to `function rect(args)`~~
  - [x] ~~rewrite all private function calls from `pub.radians()` to `radians(args)`~~
- [x] get rid of `b.go()` and `b.loop()`
- [x] instead the proposal is to have `function setup()` (for inital setup), `function draw()` (to draw your script as we used to have) and `function loop()` (to loop scripts)
  - [x] we decided against swapping `function draw()` for `function setup()`, because in > 99% of all cases, draw will be the only thing that is used in basil, it would be weird to call it setup
- [x] there will be no backwards compatibility: `b.rect()` would not work anymore
- [x] ~~users are responsible themselves not to overwrite the global functions, as this makes the code simpler and cleaner – we think if a user is advanced enough to define his own `rect()` function he should be able to figure out why the basil rect() function stopped working~~ users will get a warning when they define global functions that collide with basil functions
- [x] `b.MODEHIDDEN` etc. can be moved to a dedicated function `mode(HIDDEN)`
- [x] update all documentation to skip the `b.`
- [x] implement a `frameRate()` function to set the frame rate in looping scripts
- [x] implement saving of user defaults and resetting them in the end

That means a typical basil script would look like this in the future:

```js
// @includepath "~/Documents/;%USERPROFILE%Documents";
// @include "basiljs/basil.js";

function draw() {
  rectMode(CENTER);
  rect(50, 50, 200, 80);
  text("Hello basil 2.0!", 100, 100, 60, 60);
}
```

@basiljs Does this sound good for everybody? I would start trying to implement this in the next few days.

b-g:
👍🛠🚀

trych:
Ok, the basic switch went really smoothly and I got basil.js to work without `b.` and without `b.go()`.

[...]

PR #243, opened 2018-01-13
b-less

trych:
Here we go.

## Changes

This removes the `b.` prefix from basil, so all of basil's functions change from `b.functionName()` to `functionName()`. Constants and variables change from `b.CONSTANT` to `CONSTANT`.

The functions `b.go()` and `b.loop()` were removed from basil. Instead, `function draw()` and `function setup()` will be automatically executed now. A new `function loop()` is introduced for looping scripts.

The performance mode functionality was moved to its own function `mode()` with the possible parameters `VISIBLE`, `HIDDEN` and `SILENT`. The mode can now be changed during script execution.

A new function `frameRate()` was introduced to set the frame rate during looping scripts. Again, this frame rate can be changed during script execution.

User settings such as units, anchor point as well as default styles are saved before script execution and reset after script execution. basil scripts now run with the user's default units / the units currently set in the user's document.

All references to `b.` were removed from the inline documentation. The test script suite was modified to work with the new b-less basil, the new basil passes all of theses tests.

The `stop.jsx` script is rewritten to work with the new b-less basil.

## Notes on the implementation

The new implementation uses an anonymous function at its center just like the old one. The main difference is that after adding all of basil's functions to the `pub` object they are then injected into global space to have them available as global functions in the user scripts. The injection happens in the function `populateGlobal()` [here](https://github.com/basiljs/basil.js/blob/6463efb2e4c124b58f1cb27b8acd6dead1222c3d/src/includes/core.js#L253). In this function there is also a check, if the user is already using a function that would crash with basil's function names (example: the user has defined a `function width()` etc.). In this case, the script is cancelled with an error message.

There are two new command blocks that need to be executed before the main anonymous function takes over. Both of them are necessary to make looping scripts work correctly:

[The first one](https://github.com/basiljs/basil.js/blob/6463efb2e4c124b58f1cb27b8acd6dead1222c3d/src/main.js#L49) clears all basil functions from global space, if multiple looping scripts are executed after each other (user runs a looping script once, stops it with `stop.jsx` and then runs another (or even the same) looping script). Since the target engine `loop` keeps all the basil variables alive, there were some errors happening that were caused by the fact that the functions were saved while executing the old scripts (for example `millis()` was still depending on the old `startTime` value). While I initially thought I could work around these problems one by one, it turned out that there are too many dependencies and that the cleaner and safer way would be to just clear the basil.js variables in the beginning of the script. The clearing is only happening if the target engine `loop` is in fact used, so all scripts just using `function draw()` are not affected.

[The second one](https://github.com/basiljs/basil.js/blob/6463efb2e4c124b58f1cb27b8acd6dead1222c3d/src/main.js#L63) is necessary to load global user variables in user scripts that use setup or loop. Example:

```js
// @includepath "~/Documents/;%USERPROFILE%Documents";
// @include "basiljs/basil.js";

var counter = 0;

function loop() {
  counter++;
  println(counter);
}
```
In the old implementation this counter variable was loaded into global because the script was only executed when `b.go()` or `b.loop()` were called at the very end of the user script. In the new implementation this is not possible anymore, as there is no `b.go()` or `b.loop()`. Instead user globals need to be loaded in the very beginning. This is achieved by reading in the active user script and executing it via `app.doScript()`. While this works, this probably slows down the execution somewhat (not noticeably though), so it will only be executed in user scripts that use `function setup()` or `function loop()`. If users use only `function draw()` they have to (and really should!) move all variables into `function draw()`.

There was a `finally` clause added to the main `try catch` statement that runs commands that need to be executed in the end of the script like resetting the user settings or stating the final time.

---
@basiljs I would like to get this merged rather soon-ish, so it would be great if everyone could read over the above notes and let me know if there is something that needs to be changing.

As there are plenty of changes, a review can probably not done by looking over the code base. Instead we should depend on the test suite (it passes all tests for now) and maybe run the new basil with some special cases that we could think of (I did my best to do this already and to make sure everything works).

If there are no objections in the next few days, I then would like to go ahead and merge this, so work can continue on other functions that still need to be done for basil 2.0.

Thanks for reading! 👍


fabianmoronzirfas:
Okay. I tried a first test with the loop (that's why I asked for globals). This simple pattern does not work:  

```js
// #targetengine "loop";
// @include ~/Documents/basiljs/develop/basil.js
var counter = 0;
function setup () {

}
function loop () {
 if(counter === 10){
    println(counter);
    noLoop();
 }
}
```

fabianmoronzirfas:
Another error occurs when I try to run this script from ESTK:  
```js
// @include ~/Documents/basiljs/develop/basil.js
var counter = 0;
function setup () {

}
function draw () {
    println(counter);
}
```
There is no `app.activeScript`…

bildschirmfoto 2018-01-16 um 14 01 02

fabianmoronzirfas:
The user gets the hint 
```txt 
Running simple.js with Indesign CC.2017
### Basil Warning -> basil could not load global variables. If you need to use global variables outside of setup() and loop(), execute your script from the Extend Script Toolkit. If you are using draw(), there is no need to use setup(). Move all your global variables into draw() instead.
Running simple.js using basil.js 1.1.0 ...

[Finished in 11ms]
```
but the following script
```js
// @include ~/Documents/basiljs/develop/basil.js
var counter = 0;

function draw () {
  println(typeof counter);
}
```

    undefined


Moving all globals into `function draw` might not always be possible

trych:
@fabianmoronzirfas Thanks for giving this some testing. :) Will respond to your questions later, (at work right now), just one question for now:

> Moving all globals into function draw might not always be possible

Could you give me an example where it would not be possible? I can't think of any scenarios.

trych:
> There is no app.activeScript…

*sigh* This frickin' ESTK …

Ok, so there is two things we are discussing now:

### 1.) Loading globals
I was first totally gonna argue against your point, but after playing around with your example a bit, I think I agree that it should be possible to have global variables for a `draw()` only script. I think your example is not the best coding practice, but it is totally plausible that beginners could try something like this:

```js
// @includepath "~/Documents/;%USERPROFILE%Documents";
// @include "basiljs/basil.js";

var myPageWidth = 230;

function draw(){
  var blam = 23;
  println(doSomeCalc(11));
  println(doAnotherCalc(100));
}

function doSomeCalc (foo) {
  // do something with val here
  return foo * 2 - myPageWidth;
}

function doAnotherCalc (bah){
  // do something with val here
    return bah * 3 - myPageWidth;
}
```
I will prepare a new PR, that loads global vars in draw only scripts as well. The disadvantage is that it will make all scripts a bit slower, but I don't think it will be noticeable.

This raises one more question, and I would like to hear your opinions on that: The current implementation only loads globals that are placed before `function setup()`. See this example:

```js
// @includepath "~/Documents/;%USERPROFILE%Documents";
// @include "basiljs/basil.js";

var foo = 23;  // this variable is loaded into global

function setup () {
  // setup stuff happening here
}

function loop (){
  // basil loop happening here
}

var bar = 58;  // this var is *NOT* loaded into global
               // as it is placed after the main basil functions
```

My question is: Is this reasonable to implement it like this? Or should it load all the global variables, no matter where in the script they are?

One possible disadvantage of loading all variables into global is, that the entire source script would need to be executed before draw and the script commands that are placed on the global level (outside of `draw()`, `loop()` etc.) would be executed first. Example:

```js
```js
// @includepath "~/Documents/;%USERPROFILE%Documents";
// @include "basiljs/basil.js";

$.writeln("Print a line before draw.");

function draw (){
  println("Print a line in draw.");
}

$.writeln("Print a line after draw.");
```
would result in the console log

```
Print a line before draw.
Print a line after draw.
Print a line in draw.
```

as `function draw()`(or loop etc.) would then be executed last.

So which version does make more sense to you?
#### Version A – Loading only globals before draw or setup

```js
var foo = 23; // will be loaded
$.writeln("bar"); // will be printed

function draw() {
  // stuff
}

var baz = 58; // will *NOT* be loaded
$.writeln("bar"); // will *NOT* be printed
```

**OR**

#### Version B – Loading all globals

```js
var foo = 23; // will be loaded
$.writeln("bar"); // will be printed

function draw() {
  // stuff happening after all global stuff
}

var baz = 58; // will be loaded
$.writeln("bar"); // will be printed *BEFORE* draw
```

### 2.) app.activeScript not working in ESTK (and in default Sublime)
Ok, so I looked into this now and apparently `app.activeScript` is never defined when the ESTK is used. The problem is, we somehow need to get the user's scripts content to load the global variables. I think I found a way around the ESTK problem now and will prepare a new PR.

About Sublime: By default the app.activeScript thing does not work with Sublime's `RunInIndesign` package, because the activeScript will be the `jsRunner.jsx` that is internally used by `RunInIndesign` to start scripts from Sublime. The problem can be prevented by making one simple change in this `jsRunner.jsx` line. Change the line:

```js
var res=$.evalFile($.sblimeRunner.runFile);
```

to

```js
var res=app.doScript($.sblimeRunner.runFile);
```

I think we should prepare a PR for the package to make this its default behavior (everything else will still work the same). The problem is that vamitul, the packages developer, is not very responsive/active on GitHub. Maybe we can manage to get in touch with him over the Adobe Forums, where he is also registered and seems more active. Worst case scenario is that we do a fork and offer our own package, but let's try to get in touch with him first.

-----

@basiljs So, I will do a new PR to include globals, I would just need your opinion on which of the two options described above you guys prefer. Thanks!

fabianmoronzirfas:
>  The current implementation only loads globals that are placed before function setup()  

How is this even possible? 😳 

> Is this reasonable to implement it like this? 

No.

> Or should it load all the global variables, no matter where in the script they are?

Yes it should.

trych:
> How is this even possible? 😳

🎩 🐇

> Yes it should.

👍

trych:
@fabianmoronzirfas Made a new PR, could you give that one a quick try? This one loads all globals now, also when there is just `function draw()` and it should work in Sublime, in ESTK (even with unsaved scripts) and from the scripts panel.

fabianmoronzirfas:
Unfortunately not. This script:   
```js
// @include ~/Documents/basiljs/develop/basil.js
var counter = 0;

function draw () {
  println(typeof counter);
}
```
run from ESTK gives me this error

    stack overrun

bildschirmfoto 2018-01-17 um 13 14 34

Same thing happens when I run it from ST3

```txt
Running simple.js with Indesign CC.2017
[----------------]
  Error:InternalError: Stack overrun
    File:/Users/icke/Documents/basiljs/develop/basil.js
    Line:72
[Exited with error]

```

trych:
Could you quickly try to run your script with

```js
// @includepath "~/Documents/;%USERPROFILE%Documents";
// @include "basiljs/develop/basil.js";
```

Just want to know if that's what's causing the problem.

fabianmoronzirfas:
Yes that creates the error. 🤦‍♂️ 

How can this be?

fabianmoronzirfas:
Oh.  

```js
// @include "~/Documents/basiljs/develop/basil.js"
var counter = 0;

function draw () {
  println(typeof counter);
}
```

Works as well. leaving out the "" creates errors ……………………🤯

trych:
bildschirmfoto 2018-01-17 um 17 19 28

What, what what? Could you post your error in ASCII? :laughing:

> How can this be?

Well, it's called a bug ;) My bad, I will prepare another PR later.

trych:
Ok, this should be fixed now.
You should be able to run your scripts with globals from Sublime, from ESTK, from script panel, with quotes, or without quotes. 😃 @fabianmoronzirfas Could you do another quick test?

fabianmoronzirfas:
> What, what what? Could you post your error in ASCII? 😆

It was just a facepalm emoji  

bildschirmfoto 2018-01-18 um 11 09 37

fabianmoronzirfas:
Yes that works.  
Could you do me a big favor?  🙏 
Sketch out how the script execution currently happens? You commited so many changes that it is not that easy to follow the flow of the program. Just a simple flow chart for the execution of single shot with and without `setup` and the `loop` would be great. Maybe with some annotation on which functions execute when. So we don't have to trace everything from scratch.

trych:
Yes, that makes sense. However, I don't really know how to create a flowchart. I could sketch it by hand, that would be hard to read however. As you always know about these kind of things, do you know if there is a simple tool for that (online or Mac compatible)?

fabianmoronzirfas:
My favorite for this is a combination of graphviz and omnigraffle, but that means knowing graphviz language and the usage of omni graffle. Draw.io looks good and if you want ascii go for ascii flow.  


http://asciiflow.com/

https://www.draw.io/

https://www.graphviz.org/

https://www.omnigroup.com/omnigraffle

http://uxflow.co/

trych:
Ok, will look into this. Will probably find some time, to do this tonight.

trych:
Alright, I sketched an execution flow for [draw()](https://github.com/basiljs/basil.js/wiki/Execution-Flow-%E2%80%93-draw()) and for [loop()](https://github.com/basiljs/basil.js/wiki/Execution-Flow-%E2%80%93-loop()) and put them on the Wiki. I guess it's useful to have them around for future reference.

@fabianmoronzirfas Does this help to understand the structure?

fabianmoronzirfas:
Sort of. :-).
```js
// load global vars of the user script
var sourceScript;
try {
  app.nonExistingProperty;
} catch(e) {
  sourceScript = e.source;
}

var userScript = sourceScript.replace(/[\s\S]*[#@]\s*[i]nclude\s+.+basil\.js"*\s*;*/, "");
app.doScript(userScript);
```

So why this hack to get the script that was executed?

fabianmoronzirfas:
Why do you add all the functions also to `$.global.basilGlobal`?

trych:
> So why this hack to get the script that was executed?

This is, as far as I can see, the only way to retrieve the user script within the basil.js script execution, when it it executed from ESTK. If you know of any other way, I am more than happy to change this. :-)

> Why do you add all the functions also to $.global.basilGlobal?

This is necessary to be able to clear them out in the beginning, when they are already loaded via a target engine. I just quote myself from above:

The [clearing function] clears all basil functions from global space, if multiple looping scripts are executed after each other (user runs a looping script once, stops it with stop.jsx and then runs another (or even the same) looping script). Since the target engine loop keeps all the basil variables alive, there were some errors happening that were caused by the fact that the functions were saved while executing the old scripts (for example millis() was still depending on the old startTime value). While I initially thought I could work around these problems one by one, it turned out that there are too many dependencies and that the cleaner and safer way would be to just clear the basil.js variables in the beginning of the script. The clearing is only happening if the target engine loop is in fact used, so all scripts just using function draw() are not affected.

trych:
@fabianmoronzirfas What are your thoughts? Could it be implemented like this?

@b-g @ffd8 Have you guys had any chance to have a look at this.

I would like to get this merged soon-ish, so I can continue to work on the other things that need to be implemented. :)

fabianmoronzirfas:
Sorry. The WE was pretty busy. I can try to wrap my head around it on Friday

b-g:
Hi guys, just quickly ... I have to say that I'm a bit out. It is end of term and I have 3 more project/book launches upcoming in the next weeks ... I simply don't have time at the moment. Sorry! Please simply go ahead and don't wait on me. 

In addition I have to admit this issue becomes more and more black magic for me! 🎩 🐇 Very impressive, but not as easy to get (as even Fabian seems not to grasp everything straight away :)

My only comment is that we should make sure that we stay close enough to the standard behaviour an user expecting from Javascript e.g. global vars should also work in the middle/end of the script and not just if declared at the beginning of the script.

Peace and out unit the term is finished.

trych:
Ok, cool. No worries, as long as you don't make out any showstoppers right away, that's fine.

> My only comment is that we should make sure that we stay close enough to the standard behaviour an user expecting from Javascript e.g. global vars should also work in the middle/end of the script and not just if declared at the beginning of the script.

Yes, after consulting Fabian about it, that's how it is implemented now.

> I can try to wrap my head around it on Friday

That would be great, because I really need to continue work on these features, if I want to make the release deadline for my Todos.

Btw., as a first user feedback, I have been using the new b-less in production a bit, it seems to work just fine.

trych:
And of course, as the second part of my feedback: as I help out my students still working with Basil.js v1.x, I already forget adding the `b` all. the. time. 😀

fabianmoronzirfas:
Still to be busy for a proper review. Maybe Sunday. Sorry for that.Ome thing we should consider. 

1. Removing the anonymous self executing function

I thought I had this working already without parsing the user script. I’m wondering why it needs this step. 

Also why can’t the stop.jsx and no loop not take care of clearing the target engine so we don’t need to duplicate all the functions in global scope into `basilGlobal`? (Might be bad if the user needed to kill the script though)

trych:
> [anonymous self executing function] I’m wondering why it needs this step.

This was mainly to avoid having all private variables and functions in global space, which would again lead to potential clashes with user functions (that the user could not foresee, as they don't know how the private functions are called). The alternative would be to move all private functions and vars to a private object, but this would mean a lot of code line changes for no real gain. That's what we [discussed here](https://github.com/basiljs/basil.js/issues/241#issuecomment-355756926) and then decided to just leave the anonymous function in place.

> Also why can’t the stop.jsx and no loop not take care of clearing the target engine

By "clearing the target engine" you are probably referring to the code in [these lines](https://github.com/basiljs/basil.js/pull/243/files#diff-9ad268810f6816ce35dfa2a7d683cde0R48), right? I think it's safer to have this in the beginning of each looping script (as opposed to have it in the `noLoop()` function, as it will also work then, if the user did not previously use the `stop.jsx` script to cancel out of the previous loop script. I don't think it is a big concern though, as these code lines will only execute at all when you have a `#targetengine "loop"` in your script, so normal scripts are not affected at all.

trych:
@fabianmoronzirfas Did you have the chance to have a look on Sunday?

fabianmoronzirfas:
Nope. Sorry it is pretty busy around here

trych:
Is there any updates on this? I think I really need to continue working on this by the beginning of next week, else I cannot make the release deadlines for the ToDo's I still have listed. 😞

Fabian, is there anything specifically in the PR that needs checking or that you still have issues with?

fabianmoronzirfas:
I currently don’t find the time to do reviews. A lot to do on the family side. 

I find it wired that we have to parse the script for the include and we can't access user globals in the normal way. But I can't promise to review this in this week.

trych:
> I find it wired that we have to parse the script for the include and we can't access user globals in the normal way.

The reason the script has to be parsed, is that the entire script is already executed in the `// @include basiljs/basil.js` line. As the global variables are most likely placed after this line, the only way to retrieve them before the execution of `draw()` and `setup()` is by parsing the script beforehand.

```js
// @includepath ~/Documents/;%USERPROFILE%Documents
// @include basiljs/basil.js   // all the script is executed in this line, while myGlobalFoo has not been loaded yet, hence the parsing

var myGlobalFoo = "foo";

function draw() {
  // stuff
};
```

If you have no time for reviewing, we need to come up with another plan.
@b-g If I understand correctly, I have you go for merging, right?
@ffd8 Did you have any chance to have a look at this.

@basiljs Should I at some point just go ahead and merge? Would like to continue on other stuff, but I can't as long as this in on hold.

fabianmoronzirfas:
I think this should not be merged without reviewing and search for possible pitfalls.

trych:
I agree, but at the same time we have to move forward.

If you cannot reliably say if you have time this week, do you have any idea, when you would reliably have a few hours to have a look at this? :)

fabianmoronzirfas:
I'll guess Sunday.

trych:
Thanks, that would be great!

b-g:
Hi @trych, hi all! Sorry for the radio silence but I'm totally absorbed in the Semesterausstellungswahnsinn. Nevertheless I had some time to think about it (w/o testing the actual code) ... I no longer think that the we should remove the "go()" mechanism, as too much black magic and it breaks IMO with the global vars issue the expectations of you typically have with code which looks like standard javascript.

I guess the example below or anything using some kind of included lib file would no longer work, or?

```
// @includepath ~/Documents/;%USERPROFILE%Documents
// @include basiljs/basil.js
// @include basiljs/libraries/underscore.js // Underscore.js 1.4.4

var numbers = [1, 2, 3];

function draw() {
  _.map(numbers, ...);
};
```

fabianmoronzirfas:
> I guess the example below or anything using some kind of included lib file would no longer work, or?  
Iiiik. Yes. This is a problem.

fabianmoronzirfas:
The only solution I have is having the include at the end of the script not at the start. You can see a PoC in the microbasil repo https://github.com/fabianmoronzirfas/micro-basil/blob/master/index.js

trych:
> it breaks IMO with the global vars issue the expectations of you typically have with code which looks like standard javascript

No, global variables now *do* work as expected.
So, a user could just define his global variables as they were used to and they are then usable within the `draw()`, `setup()` and `loop()` functions.

> I guess the example below or anything using some kind of included lib file would no longer work, or?

Yes, the example would work, you simply need to change the order of the include statements:

```js
// @includepath ~/Documents/;%USERPROFILE%Documents
// @include basiljs/libraries/underscore.js // Underscore.js 1.4.4
// @include basiljs/basil.js


var numbers = [1, 2, 3];

function draw() {
  _.map(numbers, ...);
};
```

> The only solution I have is having the include at the end of the script not at the start.

While this would solve the issue, I am against this, as these include statements were always at the top and in ExtendScript or Javascript land you usually have include or require statements as the first things in your file, never as the last things.

ffd8:
@trych this is all really awesome! My own apologies for being distracted by other things.. 

I just took your branch for a very basic test run and so far so great. No issues with vars or even loading and using underscore (had to use uncompressed) – it was able to be included both above and below basiljs include statement. Was curious about the actual function of setup vs draw (if there's no looping taking place) and found that I could delete either one and the code ran fine. So perhaps it will just be a technique suggestion for what code could go into setup or just ignoring/deleting it. Custom vars and functions worked just fine regardless of position in script (above, below, between setup/draw). Scoping also worked as expected, as something invented in setup wasn't available in the draw. 

Next step will try porting some Processing (or easier yet, p5js) code to it and see how much quicker it is without any b's.... + playing with loop() mode.

Anything in particular you wanted to be tested or just try things out and see if we find issues?

Two thoughts that come up:
Should the develop branch be forked/cloned before merging? At the moment, develop fixes various bugs the master branch has (and we've all likely been using in class) without any radical changes in usage (b-less)... 

I still have to play with how the loop() mode is now working.. but would it make sense to drop loop() and have the draw() be a real draw (looping)? 

Cons:
- looping in InDesign is rare
- how would it work if executed from Sublime?
- how to easily stop (embed a key listener for esc or cmd+period to launch stop script?)

Pros:
- draw = real processing style draw
- gives more purpose to their being a setup and draw
- the game market for InDesign would explode!

trych:
@ffd8 Thanks for the feedback! Great that it seems to work well for you.

> Should the develop branch be forked/cloned before merging?

I think I would like to do one branching of the develop branch's current state with the `b` still in it, we still might need to use it in class, yes. But just to have it around for ourselves, not for release.

> would it make sense to drop loop() and have the draw() be a real draw (looping)?

We decided against it, because in 99% of cases, in basil.js you want to draw something by just running one main function *once* (as opposed to repeatedly). So what we call `draw()` should be really the function that runs once. That's why we have now `draw()` to run stuff once, `loop()` to run stuff repeatedly and `setup()` if you need to set something up before looping.

fabianmoronzirfas:
> Yes, the example would work, ~~you simply need to change the order of the include statements~~ (Edit: Actually now that I think about it, it works no matter the order of the two include statements, see also Ted's comment below, so the example that you posted should work just fine).

I tried a simple example with my own function in a additional script and it didn't work. :-/ I'll need to investigate this further.

Two things that came to my mind tonight (that's why I'm writing at 4:30 in the morning):  

1. The execution of the script by using `go()`/`b.go()` or not has nothing to do with beeing `b.less`. Might be useful to rework this into two PR branches to have one where the functions get all pushed into global scope and another that looks at the execution of the script with `go` or without `go`.  
2. Another fancy way to execute functions right away in Javascript is the keyword `void`. Since our `setup` and `draw` don't newer return something it could be a way to make it self-executable.  

This is valid JS code:  

```js
 var g = 'bah';
 void function setup(){
        $.writeln(g);
        g = 100;
    }()


 void function draw(){
        $.writeln(g);
    }()

```

I currently have no idea on how we could get a loop to execute itself over and over, but this might be done in the targetengine or something.

fabianmoronzirfas:
So I have this setup in the basil folder on the latest state of the PR branch:  

```plain
.
├── basil.js
(…)
├── test
|   (…)
│   ├── foo.jsx
(…)
│   ├── index.jsx
(…)
```



index.js includes foo.jsx and basil.js


```js
// @include "../basil.js"
// @include "./foo.jsx"


function draw() {
  $.writeln(bah(10));
};
```


This is foo.jsx

```js
function bah(x){
  return x * 2;
}
```

And the script stops for mo at line 72 in basil.js when executed from ESTK\*
bildschirmfoto 2018-02-08 um 04 49 42

It works when executed from Sublime Text and from within the Scripts Panel from InDesign. Even though ESTK might be a thing of the past we can't ignore it. 

\* That's where my iiiikk came from.

trych:
> The execution of the script by using go()/b.go() or not has nothing to do with beeing b.less. Might be useful to rework this into two PR branches to have one where the functions get all pushed into global scope and another that looks at the execution of the script with go or without go.

While I agree that this might have been a possibility to approach this initially, I doubt that this would have speeded up review processes. And right now this would just mean a whole lot of re-work with little gain.

b-g:
@trych OK + Sorry + Convinced now + Please go ahead from my end! 🙏

fabianmoronzirfas:
For me this is nebulous. I don't see the logic behind it.

```js
// @include ./foo.jsx
// @include ../basil.js
```

or

```js
// @include ../basil.js
// @include ~/Documents/basiljs/test/foo.jsx
```

or

```js
// @includepath ~/Documents/basiljs/test/
// @include ../basil.js
// @include foo.jsx
```

fabianmoronzirfas:
#Things Noted:

## Deleting global vars in main.js

I don't know how this can work when calling `basilGlobal.length` or I don't get it.  

```js
if($.engineName === "loop" && $.global.basilGlobal) {
  // 
  for (var i = $.global.basilGlobal.length - 1; i >= 0; i--) { //$.global.basilGlobal.length not basilGlobal.length ? 
    if($.global.hasOwnProperty($.global.basilGlobal[i])) { // the same here ?
      try{
        delete $.global[basilGlobal[i]]; // and here as well ?
      } catch(e) {
        // could not delete
      }
    }
  }
  delete $.global.basilGlobal;
}
```

## Loop:  

When using the loop I get the output that the execution is finished before the script actually executes.  


## Basil Include:  

When I write the include statement in `// @include '/Users/me/Documents/basiljs/develop/basil.js'` I get this error.


```plain
Running index.jsx with Indesign CC.2017
[----------------]
  Error:SyntaxError: Unterminated string constant
    File:/Users/icke/Documents/basiljs/develop/basil.js
    Line:72
[Exited with error]
```


Regex should be maybe like this (no garantee):


    ^.*?[\/\/.*?@|\#].*?include(?!path)["|']?(.+)["|']?

bildschirmfoto 2018-02-11 um 15 48 02


## Global functions

Shouldn't  the functions in `src/includes/global-functions.js` be

    var forEach = function()…

or

    function forEach ()…

same goes for Hashlist.

It would be great if @ffd8 could stresstest the loop a little more. 

----

About the self execution of the script. It still is blackmagic that is happing but actually I nice workaround. 👍 @trych 

I guess 

```js
// @include "../basil.js"
// @include "./foo.jsx"
```

Does not work because the executing script is `basil.js` not my `index.js` anymore. Wich is a bummer. I also got some random glitches that I could not reproduce yet. 

- Some stack overrun when running the main.js directly.
- Seldom wired behavior with loop which should be stress tested with some more complex examples that might be in @ffd8 s drawer.

Another question that I have:

- What happens with a very large userScript. It needs to be catched by the try catch block for regex analysis to replace the basil include statement?
- tbd

I still don't think we should merge yet. This needs some more testing but in general I think this could be a way to go.\* I see no other possibility (yet) to make it self executable. Except for putting the include for basil at the end. Which is not nice but still a valid solution.  

\* @trych If it itches you to keep working on your other features it might be good to do so on a branch that is forked from b-less. If we have the feeling that b-less is ready for merge, the other MRs can go first and then we merge b-less into develop.  

I'm out for today.  Sorry for all the delay in reviewing this. I have a lot going on with my family. This might be solved in 2 weeks. Then I get some more spare time to fiddle on basil again. (e.g. Website)

src/main.js
```diff
@@ -43,40 +43,63 @@
...
-  // @include "includes/ui.js";
-
-  init();
-
-})(this, app);
 
+// clearing global space if it is still populated from previous run of a loop script
+// to ensure basil methods work properly
+if($.engineName === "loop" && $.global.basilGlobal) {
+  for (var i = basilGlobal.length - 1; i >= 0; i--) {
```
fabianmoronzirfas: shouldn't this  `basilGlobal[i]` not be `$.global.basilGlobal[i]` see my latest comment in discussion

trych:
@fabianmoronzirfas Thanks for the review!
(sorry, I have been busy myself this week, hence the belated answer)

About the points you raised:

---

> I don't know how this can work when calling basilGlobal.length or I don't get it.

What exactly do you mean? That the var is called `$.global.basilGlobal` instead of just `basilGlobal`? It's basically the same thing, but you are right, for consistency, it should be renamed to just `basilGlobal`. Or do you mean something else?

---

> When using the loop I get the output that the execution is finished before the script actually executes.

Yes, this should be fixed, will put it on my todo.

---

> When I write the include statement in // @include '/Users/me/Documents/basiljs/develop/basil.js' I get this error.

Will look into this.

---

> Shouldn't the functions in src/includes/global-functions.js be `var forEach = function()…`

`filter()`, `map()`, `forEach()`:
Those functions were special functions in the "old" basil in that they were no `b.` methods, but global functions (see [this use of forEach in an example script](https://github.com/basiljs/examples/blob/master/input/loadStrings.jsx#L28)). I guess now, since all functions are global functions those functions should be treated the same, yes.

`HashList`:
This was a class in the old basil, should it just be treated the same as all global functions then?

---

> About the self execution of the script. It still is blackmagic that is happing but actually I nice workaround. 👍 @trych

Thanks! 😃 

---

> I guess `// @include "../basil.js"...` Does not work because the executing script is basil.js not my index.js anymore.

Something along those lines. Also, the reason why some scripts with relative paths work from the script panel and from Sublime, but not from ESTK is that when you run a script from outside ESTK it generally adds the location of the user script to your include paths (`$.includePath`). It does *not* do that, when you run from ESTK, that's why you get some errors when using relative paths from ESTK.

My suggestion therefore is to have a small section on the website named "Extending basil.js", where we explain how to include external scripts. Once users have this as a pointer, it should not be a problem anymore. I could help write that section, if there is the need.

---

> Some stack overrun when running the main.js directly.

That is to be expected, but should be a non-issue, as users never will get to see a main.js file. As a basil dev, don't run main.js 😉 (and yes, it happened to me, too).

---

> Seldom wired behavior with loop which should be stress tested with some more complex examples that might be in @ffd8 s drawer.

@ffd8 Yes, would be nice, if you could throw some of your loop scripts at the new basil. :)

---

> What happens with a very large userScript. It needs to be catched by the try catch block for regex analysis to replace the basil include statement?

Well, it's not the script that you catch, but the tiny error that you have a non-existing property, no matter which size your script has. Then the script is loaded into the `e.source` property. I don't think large scripts should cause any problems as there are lots of huge strings all over the place in the InDesign Object Model. Then the regex just clips of the first few lines basically, so script size should not really matter for the regex either.

Does somebody have a huge basil script, so I could quickly give this a test run?

---

> If it itches you to keep working on your other features it might be good to do so on a branch that is forked from b-less.

Yes, I thought of that, but I just needed a confirmation first, if the general direction is alright. Now that I have that from you, I am fine with just branching of `b-less`.

---

>  Sorry for all the delay in reviewing this. I have a lot going on with my family. This might be solved in 2 weeks.

No worries, I know how tricky time management gets with kids around. So thanks for doing the review.

---

> This needs some more testing but in general

My suggestion: As this passes the entire test suite already and my custom tests, @ffd8 tests and I am using it without any issues in production already: How about I convert all the example scripts to `b-less` (which we need to do anyways) and if they all work, then we could consider this tested (+ some more loop tests by @ffd8 would be nice). Would that be ok for everyone?

The points above I will put on my todo list, so I keep on working on this now. :)

fabianmoronzirfas:
Hi @trych I need to set this onto my todo. Can't work this through tonight but just a comments.  

About the `basilGlobal`: Of course after some thinking you can call `$.global.basilGlobal` or `basilGlobal` we should stick to just one way to call it or it creates confusion.

src/main.js
trych: Ok, now I looked into this issue. I would leave all global variables as `basilGlobal` if possible. The *first* instance of `basilGlobal` (in the if expression) I cannot just leave as `basilGlobal`, because that throws an error:

```
basilGlobal ist undefiniert
```

That's why I put it as `$.global.basilGlobal` there. However, I probably could test for this instead:

```js
if($.engineName === "loop" && $.global.hasOwnProperty("basilGlobal") { ... }
```

Do you think that's clearer?
Still, the delete statement in the last line of this block needs to be

```js
delete $.global.basilGlobal;
```

as the `delete` statement can only delete properties.

trych:
> When I write the include statement in // @include '/Users/me/Documents/basiljs/develop/basil.js' I get this error.

> When using the loop I get the output that the execution is finished before the script actually executes.

@basiljs Those two are fixed now.

trych:
Ok, talked to @b-g and @ffd8 yesterday, I will merge this later today.

Before the merge there will be a final branch off the `develop` branch, that we name `develop-v1`, which is for those people that for some reason need the most feature rich basil1 version in the future. I'll add two lines in the `develop` README, so people will be able to find this version.

trych:
Ok, first step done, the frozen v1 branch can be found [here](https://github.com/basiljs/basil.js/tree/develop-v1).

trych:
We reached the end of the coding sprint aaaaaaaand ...

we're officially b-less!

Goodbye good ol' `b.`, it has been a hell of a ride. I remember, when I first met you in 2013, I had no idea what a variable was. 😆

🎉 😄 🎉 😢 🎉 😄 🎉

PR #250, opened 2018-02-25
b-less implementation improvements

trych:
This is a slightly tweaked alternative to the `b-less` implementation that I put up for PR a few weeks ago.

The main difference is that it avoids the regex that parses the source script file, as I think this was a bit to unsafe. It might have failed on special cases like this:

```js
if(something) {
  // @include basiljs/basil.js
  // some comment that makes the regex fail
}
```

The new implementation avoids the regex by setting a global variable `basilRunning` to `true` the first time an execution chain enters `basil.js`. If `basil.js` is entered again (through the execution of the source script to load the global variables outside of `draw()`), it cancels right out again, because basil is already running.

So there is no need anymore to take the `// @include basil/basil.js` lines out with a regex.

Other improvements:

- basil scripts can be chained now from a main script (which is something that was not possible with the old basil).
- main.js does not create a stack overflow anymore when you run it by accident ;)

@fabianmoronzirfas (or anyone from @basiljs) If you agree that this tweak improves the previous implementation, we could pull it into the `b-less` branch. :smile:

This still passes the entire test suite.

### DIFF
--- a/basil.js
+++ b/basil.js
@@ -45,45 +45,49 @@
 // @target "InDesign";
 
 
-// clearing global space if it is still populated from previous run of a loop script
-// to ensure basil methods work properly
-if($.engineName === "loop" && $.global.basilGlobal) {
-  for (var i = basilGlobal.length - 1; i >= 0; i--) {
-    if($.global.hasOwnProperty(basilGlobal[i])) {
-      try{
-        delete $.global[basilGlobal[i]];
-      } catch(e) {
-        // could not delete
+if(!$.global.hasOwnProperty("basilRunning")) {
+
+  $.global.basilRunning = true;
+
+  // clearing global space if it is still populated from previous run of a loop script
+  // to ensure basil methods work properly
+  if($.global.basilGlobal) {
+    for (var i = basilGlobal.length - 1; i >= 0; i--) {
+      if($.global.hasOwnProperty(basilGlobal[i])) {
+        try{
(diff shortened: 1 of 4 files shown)


fabianmoronzirfas:
This sounds great. I'll try to make some time for testing tonight. One thing that came to my mind. What happens to the `basilRunning` in a loop that gets stopped by killing a ESTK process?

trych:
> What happens to the basilRunning in a loop that gets stopped by killing a ESTK process?

That will cause problems. (As this did cause problems with the old basil as well, I think. I still would label this whole loop business experimental as it always has been labled and tell the user that they *have* to use `stop.jsx` to quit their looping scripts.

However, we could maybe set `basilRunning` to `false` again as soon as a `function loop()` is detected and then set a var like `basilLooping` to `true` instead or something like that. Will have a look at this later today.

trych:
Ok, I just placed the deletion of the prop (now renamed to `basilTopLevelExecution`) at the earliest possible point (right after all potential recursive calls), so it gets deleted before the `loop()` (or `draw()`, or `setup()`) function is even run. That way it is safely deleted in case a loop script is force quit. Could you give this some testing, Fabian?

I think this is a lot safer compared to the implementation using the regex.

fabianmoronzirfas:
I did not make it yesterday and wont make it tonight. Friday might work

PR #254, opened 2018-03-08
canvasMode() for 3+ page spreads

ffd8:
our FACING_PAGES, FACING_MARGINS, FACING_BLEEDS was hard coded for a 2 page spread, but now works with 3+ spreads.


ffd8:
found another bug.. pulling combined branch

PR #255, opened 2018-03-08
canvasMode() bugs fixed for 3+ page spreads and single page on facing_pages

ffd8:
- FACING_PAGES, FACING_MARGINS, FACING_BLEEDS was hard coded for a 2 page spread, but now works with 3+ spreads.

- fixes bug that BLEED, FACING_BLEEDS, FACING_MARGINS all had issues if used on single page within a document with facing_pages - could easily happen if starting a new document with only one page in that mode

### DIFF
--- a/basil.js
+++ b/basil.js
@@ -1248,6 +1248,7 @@ var updatePublicPageSizeVars = function () {
   var pageBounds = currentPage().bounds; // [y1, x1, y2, x2]
   var facingPages = currDoc.documentPreferences.facingPages;
   var singlePageMode = false;
+  var spreadCount = currentPage().parent.pages.length;
 
   var widthOffset = 0;
   var heightOffset = 0;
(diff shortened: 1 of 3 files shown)


trych:
Thanks!

Had a quick look, this one still has issues if there are spreads with differently sized pages.
Also, this does not calculate the width and offset in FACING_BLEEDS and FACING_MARGINS correctly, if margins or bleeds are unequal.

I also think the updatePublicPageSizeVars() function could be cleaned up quite a bit. I think I will give this one a stab, would be good to have this fixed finally. Will branch of this once more, to not mess anything up.

trych:
@ffd8 As we discussed yesterday, this was now fixed [here](https://github.com/basiljs/basil.js/pull/257) and now works reliably when margins and bleed is set up in funky ways.

Will close this PR.

Issue #166, opened 2017-02-07
b.bounds() issues


trych:
Hi there,

I found a few problems with `b.bounds()` and have put together some improvements that could be made to the function.

Here we go:

## 1) Incorrect bounds on rotated frames

At the moment the results of `b.bounds()` on rotated text frames are incorrect, as values like endBaseline are not taken into consideration. Fortunately properties like `left` or `right` are all relative positions (as opposed to `x` and `y`), so I would suggest to calculate the bounds in a way that rotates along the rotated word. That way we can always use a line like

```js
var rect = b.rect(bounds.left, bounds.top, bounds.width, bounds.height);
```

to draw a rect around a text (if the matrix is still rotated), no matter if that text is rotated or upside down. I have already implemented this in a branch I am working on, and can report that it works. Keep in mind that this does _not_ change the behavior for non-rotated frames at all, it only makes it work for rotated frames, too!

## 2) Text collections throw an error

Text collections like `[Object Words]` until yesterday's commit threw the error

```js
"b.bounds(), invalide type of parameter! Can't get bounds for this object."
```

as they were not recognized as a text object by basil. Now they are and `b.bounds()` throws an error about the fact that baseline is not a property. Therefore we need to address those collections and pick their `baseline`, `endBaseline` from the objects contained in the collections.

## 3) Text object throws yet another error

Text objects (as in `[Object Text]`) that are created by using the `itemByRange()` method throw a different error. They _do_ have a `baseline` property, but for some reason this property is an array whose only value contains the actual `baseline` value. Why on earth that is the case, I have no idea. But we need to address this type of object seperately yet again.

## 4) Calculate actual bounds of multiline text
At the moment something like

```js
var tf = b.text(b.LOREM, 0, 0, 200, 200);
var bounds = b.bounds(tf.paragraphs.firstItem());
b.rect(bounds.left, bounds.top, bounds.width, bounds.height);
```

draws really incorrect bounds (see image). For multiline text the actual last line is not taken into consideration. Also the right edge of the paragraph is calculated by the last line when it should be calculated by the longest line. I suggest to treat every type of text that the actual bounds are calculated (in this for example by figuring out the longest line of the paragraph).

![bildschirmfoto 2017-02-07 um 12 25 10](https://cloud.githubusercontent.com/assets/9803905/22689801/e6f89ffa-ed32-11e6-8921-5a3461b6e796.png)

## 5) Smarter handling of `disconnected` text parts
[...]

PR #256, opened 2018-03-08
canvasMode(BLEED/MARGIN/FACING_BLEEDS/FACING_MARGINS) fix for bounds()

ffd8:
The bounds was ignoring our canvasMode() - a potential bug since a long time that was occasionally run into.


fabianmoronzirfas:
What about this PR? Is it still valid? Should get a review

fabianmoronzirfas:
@basiljs

ffd8:
I believe @trych ended up null+voiding this request when he did an overhaul recently on bounds? (canvasMode is still broken in current develop branch). I think his branch still needs to be added.. so I leave it up to him if this should be merged or not.

trych:
This is still valid, but needs some re-work, as we now have global private variables to keep track of the x and y origin (as determined by the currently set canvas mode and page). Will have a look at this soon.

trych:
Ok, I implemented `currOriginX` and `currOriginY` into the function, so it is more reliable now. Note that this does not fix the issues described in #166 yet, fixing those will require a major rewrite of the function and can hopefully done some time in the future.

Will merge.

ffd8:
We've discovered that `canvasMode(MARGIN)` (and `FACING_MARGINS`) still messes up the `bounds()`.
Luckily no problems with normal/facing BLEED or PAGES

 ```
canvasMode(MARGIN); // FACING_MARGINS

  var r = rect(0, 0, width, height);

  var bnds = bounds(r);
  
  fill(255, 0, 0);
  var r2 = rect(bnds.left, bnds.top, bnds.width, bnds.height);
```

Issue #100, opened 2016-09-26
Convert text to path  


fabianmoronzirfas:
This is a easy one. See http://yearbook.github.io/esdocs/#/InDesign/Paragraph/createOutlines

could be something like:  

    /**
     * @param Story|TextFrame||Paragraph||Line||Word||Character
     * @param deleteOriginal default to true
     * return polygon group
     */
    var poly = b.outline(item, [deleteOriginal]);
    
    
    pub.outline = function(item, deleteOriginal) {
      if(deleteOriginal === null || deleteOriginal === undef){
        deleteOriginal = true;
      }
      var res = null;
      if(item instanceof Story||TextFrame||Paragraph||Line||Word||Character){
        res = item.createOutlines(deleteOriginal);
      }else{
        warn();
      }
    return res
    }

trych:
Nice idea, would like to see this implemented.

One thing though: I *think* the deleteOriginal parameter is not really consistent with how basil.js parameters work. There are no functions in basil which take a `true` or `false` as argument (edit: `b.close()` seems to be an exception?). I'm not sure myself, why that is, is this maybe adopted from Processing?

For example when I was implementing `b.gradient()` I thought I would allow a boolean argument that makes the gradient radial if true. I then settled for the (what I thought is the) more basil like way of introducing `b.LINEAR` and `b.RADIAL` instead.

Can somebody tell me – or can we together decide – which one is the preferred way of doing this?

fabianmoronzirfas:
Hm. I get your point but what would be the CONSTANT to use? B.DELETEORIGINAL or B.KEEPORIGINAL? I think in a binary case like this one (yes vs no) keeping the original object it is okay to use a boolean. The gradiant could also be a b.WAVE gradiant. I know this does not exist but the possibility of another type is there. In the UI it is a dropdown not a checkbox.

trych:
Yes, I see what you mean. I'm not really against it, I just wanted to point out that these boolean arguments have not really been part of basil.js so far. Maybe @b-g knows if there is any specific reason for that or if there just never was any need for it?

ffd8:
True this does stem back to some aspects of Processing, ie. endShape(CLOSE). And for the gradient example you designed, I think that's the right way to go, having a default type, but allowing one to specify another (why should 'true' == radial ?). 

[...]

PR #268, opened 2018-04-02
basiljs gets points!

ffd8:
- 4 new point based functions: `createOutlines()`, `addPoints()`, `points()`, `pointsInPaths()`
- w/ initial research by @kajetansom on point getting and interpolating, extended to handle live type, multi-line type, pageitems, etc.
- `createOutlines()` built upon outline pseudocode from @fabianmoronzirfas on issue #100 

Some sample code for the new functions:

```js
  // create text and outline it
  textSize(100);
  var tf = createOutlines(text("basil.js", 50, 50, width, height));
  
  // add # of points between points of outline
  addPoints(tf, 5);
  
  // get all points as 1D array
  var pts = points(tf);
  
  // get all points in separate paths as 2D array
  var ptsPaths = pointsInPaths(tf);
  
  
  // redraw 1D array from points()
  tf.remove();
  noFill();
  beginShape();
  for (var i = 0; i < pts.length; i++) {
      vertex(pts[i].x, pts[i].y);
  }
  endShape();
  
  
  // redraw 2D array from pointsInPaths()
  //tf.remove();
  noFill();
  translate(0, 150);
  for (var j = 0; j < ptsPaths.length; j++) {
    beginShape(CLOSE);
    for (var i = 0; i < ptsPaths[j].length; i++) {
        vertex(ptsPaths[j][i].x, ptsPaths[j][i].y);
    }
    endShape();
  }


  // redraw wildstyle
  noFill();
  translate(0, 150);
  var offset = 15;
  for (var j = 0; j < ptsPaths.length; j++) {
    beginShape(CLOSE);
    for (var i = 0; i < ptsPaths[j].length-offset; i++) {
        vertex(ptsPaths[j][i+offset].x, ptsPaths[j][i+offset].y);
        vertex(ptsPaths[j][i].x, ptsPaths[j][i].y);
    }
    endShape();
  }
```
screen shot 2018-04-02 at 20 33 30

### DIFF
--- a/basil.js
+++ b/basil.js
@@ -5020,6 +5020,78 @@ pub.duplicate = function(item) {
 
 };
 
+/**
+ * @description Add additional anchor points to any page item. 
+ * Particularly useful with curved items in combination with <code>points()</code> and <code>pointsInPaths()</code>.
+ *
+ * @cat Document
+ * @subcat Transformation
+ * @method addPoints
+ * @param  {PageItem|Text} The page item to add additional anchor points to.
+ * @param  {number} Amount of additional anchor points to add between points.
+ */
+pub.addPoints = function(obj, numPoints) {
+  var amount = 1 / numPoints;
+  var prevPoint;
+  
+  try {
+    if(obj.polygons.length === 0) {
+      for (var i = 0; i < obj.paths.length; i++) {
+        pushPoints(obj.paths.item(i));
+      }
+    } else {
+      for (var k = 0; k < obj.polygons.length; k++) {
+        var paths = obj.polygons[k].paths;
+        for (var i = 0; i < paths.length; i++) {
+          pushPoints(paths.item(i));
+        }
+      }
+    }
+  } catch(e) {
+    error("addPoints(), invalid pageItem. Maybe sent grouped items, send one at a time.");
+  }
+
+  function pushPoints(pathObj) {
+    var myNewPath = [];
+    var points = pathObj.pathPoints;
+
+    for (var c = 0; c < points.length; c++) {
+      var point = points.item(c);
+      if(c === 0) {
+        prevPoint = points.item(points.length-1);
+      }
+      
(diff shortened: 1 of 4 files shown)


fabianmoronzirfas:
Nice. I'm a bit behind on my reviews lately. Maybe @trych finds some time soonisher then me.

trych:
Sweet! Can't make any promises, but will try to have a look at it soon!

fabianmoronzirfas:
Did you see these conflicts? for math.js? 
Did you rebuild the basil.js file?

trych:
> Try/Catch, what is the problem with them?

It's just not very clean coding and a bit like "let the user do their stuff and see what happens" instead of planning it carefully and do proper checking of inputs. And as it stands right now (without the native InDesign error message) it is even more confusing, because I might get a error message like `Be sure to use:\n Story | TextFrame | Paragraph | Line | Word | Character"` after passing a word, so that's confusing. Mixing this with a native error message is also not good, because then you might have different local languages mixed in one error message.

Instead I would – as it is done in most basil functions currently – carefully check by code for each condition that needs to be fulfilled to use the function. That means first check, if the correct type of object was passed. If not give a helpful basil error. This error should avoid most further errors right away. Then, if necessary, check if the object has a path (or whatever your function needs). If not, give a helpful basil message. And so on.

Right now there are quite a few ways to "break" the functions and get caught in the try/catch statement. If there is proper check of the inputs, that could be avoided. Does that explanation help? If not, let me know, I might find some time to put together an example for one of the functions.

> do I need to cancel this pull, or can I simply modify it?

I think the cleanest would be to actually cancel this PR and prepare a new one, else all these formatting changes back and forth would end up in the git history. Would that be okay for you?

ffd8:
Closed! Replaced by a much better set of functions in #352 – integrating the helpful suggestions above. Tested it against so many scenarios.. should be quite safe now... and no try/catch!

Issue #266, opened 2018-03-28
Consistent order of jsdoc attributes


trych:
Before I reapply new categories to the basil functions (see #264), I would like to make the order of attributes of the inline jsdoc documentation consistent. At the moment, it is somewhat mixed and inconsistent. Here is my suggestion of how the order and appearance should look:

```
/**
 * @description The description.
 * @method methodName
 * @cat Category
 * @subcat Subcategory, if any
 * @param  {parameterType} param Some parameter, if any.
 * @return {returnType} The return value, if any.
 *
 * @example <caption>Example of using the method, if any</caption>
 * myMethodCode();
 */
```
Some notes:

- As we need the `@description` tag a few times (for using inline html) I would suggest to use it *all the time*, to be consistent.
- Is it okay to always have the description in one line? I find it difficult to edit the descriptions with the manual line breaks.
- Should we leave the `@method` tag at `@method` or change it to `@function` (as we have functions now, not b-methods anymore)? p5.js has it at `@method` although they use functions.

Does this look okay? @fabianmoronzirfas If so, I would go ahead, and edit them all to be consistent.

fabianmoronzirfas:
++ for `@description` all the time
+- for `@description` in several lines. Might reduce readability when user has no line-wrap enabled
-- for change of `@method` if not tested with output. I specifically use tags generated from documentation.js I can't tell if this introduces a breaking change. I will need to test this first.

trych:
> Might reduce readability when user has no line-wrap enabled
Yes, I find it just super annoying when you have to add a few words and then need to shift around the twenty following lines manually. Or is there some Sublime package that I am missing out on?
What do the other ones think on that point? @basiljs 

`@method` will stay in place then, no problem.

@fabianmoronzirfas, is the order also alright? I mean I guess it does not make a difference for parsing at all, I just want a consistent order that reads well for the devs.

fabianmoronzirfas:
The order is fine for me. You can toggle the soft wrap in st3. Is a build in setting

trych:
@fabianmoronzirfas Could we turn on [Markdown for JSDoc](http://usejsdoc.org/plugins-markdown.html)?

Once we do that we can could write inline code in the descriptions like this:

```
[...] my description containing a `code snippet` that is easy to type [...]
```

instead of the current:

```
[...] my description containing a <code>code snippet</code> that is harder to type [...]
```

And we could also have easier typing of links etc.

Or is this already turned on?

[...]

PR #270, opened 2018-04-13
Consistent formatting of jsdocs comments

trych:
This makes the formatting of all jsdocs inline documentation consistent and therefore cleans it up considerably. After quite a bit of research and testing around, I came up with a formatting style that looks clean, is maintainable and works with the doc generation (once we have the markdown rendering in place).

Here is a before/after of the `margins()` function

Before
```js
/**
 * Sets the margins of a given page. If 1 value is given, all 4 sides are set equally. If 4 values are given, the current page will be adjusted. Adding a 5th value will set the margin of a given page. Calling the function without any values, will return the margins for the current page.
 *
 * @cat Document
 * @subcat Page
 * @method margins
 * @param {Number} [top] Top margin or all if only one.
 * @param {Number} [right] Right margin.
 * @param {Number} [bottom] Bottom margin.
 * @param {Number} [left] Left margin.
 * @param {Number} [pageNumber] Sets margins to selected page, currentPage() if left blank.
 * @return {Object} Current page margins with the properties: top, right, bottom, left.
 */
```

After
```js
/**
 * @description Sets the margins of a given page. If 1 value is given, all 4
 *          sides are set equally. If 4 values are given, the current page will
 *          be adjusted. Adding a 5th value will set the margin of a given page.
 *          Calling the function without any values, will return the margins for
 *          the current page.
 *
 * @cat     Document
 * @subcat  Page
 * @method  margins
 *
 * @param   {Number} [top] Top margin or all if only one.
 * @param   {Number} [right] Right margin.
 * @param   {Number} [bottom] Bottom margin.
 * @param   {Number} [left] Left margin.
 * @param   {Number} [pageNumber] Sets margins to selected page, currentPage()
 *          if left blank.
 * @return  {Object} Current page margins with the properties: `top`, `right`,
 *          `bottom`, `left`.
 */
```

As you can see, this now also strictly stays within the 80-characters line limit to be better readable in most editors.

The syntax looks like this now:
```js
/**
 * @description The description of the method. If the description is longer, it
 *          wraps around to an indentation value that aligns with other
 *          information below.
 *
 * @cat     Category
 * @subcat  Sub-Category
 * @method  methodName
 *
 * @param   {Type} name A parameter.
 * @param   {Type} name A parameter with a description that is long enough that
 *          it wraps into the next line.
 * @param   {Type} [name] An optional parameter.
 * @return  {Type} Description of the return value. This also wraps into the
 *          next line if the description is long enough.
 *          
 * @example <caption>How to use this function</caption>
 * var foo = methodName(); // code lines must not use indentation
 */
```

All methods have been changed to adhere to this new syntax. Markdown was added for inline code snippets and for lists.

### DIFF
--- a/basil.js
+++ b/basil.js
@@ -523,16 +523,18 @@ var ERROR_PREFIX = "\nBasil.js Error -> ",
 // ----------------------------------------
 
 /**
- * System variable which stores the width of the current page.
- * @property {Number} width Width of the current page.
+ * @description System variable which stores the width of the current page.
+ *
  * @cat Environment
+ * @property {Number} width Width of the current page.
  */
 pub.width = null;
 
 /**
- * System variable which stores the height of the current page.
- * @property {Number} height Height of the current page.
+ * @description System variable which stores the height of the current page.
+ *
  * @cat Environment
+ * @property {Number} height Height of the current page.
  */
 pub.height = null;
 
@@ -588,7 +590,8 @@ var addToStoryCache = null, /* tmp cache, see addToStroy(), via InDesign externa
 
 /**
  * @description The <a href="https://developer.mozilla.org/en/JavaScript/Reference/Global_Objects/Array/filter">filter()</a> method creates a new array with all elements that pass the test implemented by the provided function.
- * @cat Data
+ *
+ * @cat    Data
  * @subcat Array
  * @method Array.filter
  * @param  {Function} callback The Function is a predicate, to test each element of the array. Return true to keep the element, false otherwise.
@@ -614,7 +617,8 @@ if (!Array.prototype.filter) {
 
 /**
  * @description The <a href="https://developer.mozilla.org/en-US/docs/JavaScript/Reference/Global_Objects/Array/map">map()</a> method creates a new array with the results of calling a provided function on every element in this array.
- * @cat Data
+ *
+ * @cat    Data
  * @subcat Array
  * @method Array.map
  * @param  {Function} callback Function that produces an element of the new Array.
@@ -650,13 +654,13 @@ if (!Array.prototype.map) {
 }
 
 /**
-* Used to run a function on all elements of an array. Please note the existance of the convenience methods <code>stories()</code>, <code>paragraphs()</code>, <code>lines()</code>, <code>words()</code> and <code>characters()</code> that are used to iterate through all instances of the given type in the given document.
+* @description Used to run a function on all elements of an array. Please note the existance of the convenience methods <code>stories()</code>, <code>paragraphs()</code>, <code>lines()</code>, <code>words()</code> and <code>characters()</code> that are used to iterate through all instances of the given type in the given document.
 *
-* @cat Data
+* @cat    Data
 * @subcat Array
 * @method Array.forEach
-* @param {Array} collection The array to be processed.
-* @param {Function} cb The function that will be called on each element. The call will be like function(item,i) where i is the current index of the item within the array.
+* @param  {Array} collection The array to be processed.
+* @param  {Function} cb The function that will be called on each element. The call will be like function(item,i) where i is the current index of the item within the array.
 */
 forEach = function(collection, cb) {
   for (var i = 0, len = collection.length; i < len; i++) {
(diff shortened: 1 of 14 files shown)


trych:
@basiljs Does anyone have any comments on this? Else I'll just go ahead and merge it.

fabianmoronzirfas:
Great that you cleaned that up. I don't like manual indent of the description. It could be interpreted as Markdown code blocks if we parse the description.  

+1000 for the cleanup
-- on the manual indent

trych:
It's not not interpreted as markdown code blocks during parsing, I tested this quite intensely. It parses just fine.

trych:
(The scenario you describe would only happen, if there were some blank lines in between blocks, which is never the case)

trych:
There is of course a way to [create and update this indentation automatically](https://github.com/20Tauri/DoxyDoxygen) in Sublime.  I could do a write up in the Wiki, about how to set up the formatting.

Would that be okay for you?

trych:
Also, the reason I went with indentation in the first place is, because I did some research about what the standard is of how to format such jsdoc matter and found this thread: [What is the proper/canonical formatting of long JSDoc lines?](https://stackoverflow.com/questions/30676286/what-is-the-proper-canonical-formatting-of-long-jsdoc-lines)

The accepted answer there points out that even Google recommends in its JavaScript styleguides to indent lines after the first one. So I thought this seems very reasonable.

fabianmoronzirfas:
>There is of course a way to create and update this indentation automatically in Sublime. I could do a write up in the Wiki, about how to set up the package.

>Would that be okay for you

I'm actually not using SublimeText anymore. 

I guess it is not about making Fabian happy. @basiljs had anyone else an opinion or is it just me complaining?

fabianmoronzirfas:
FYI 
8483151c-6add-4a0d-ae50-a3485ecd2f53

fabianmoronzirfas:
The display depends very much on the size of the window.

trych:
> I'm actually not using SublimeText anymore.

But how do you run ExtendScript? :D

> The display depends very much on the size of the window.

Yes, but that it does not render well on mobile phones is not really an argument, I think. If it were about that, we should also do without the 80-character line limit and it would look much better. But nobody is editing basil.js source code on their mobile? Or are they?

I too would like to hear the opinions of the others. @basiljs ?

And @fabianmoronzirfas, what alternative do you actually prefer? A or B?

A)

```js
/**
 * @description Sets the margins of a given page. If 1 value is given, all 4
 * sides are set equally. If 4 values are given, the current page will be
 * adjusted. Adding a 5th value will set the margin of a given page. Calling the
 * function without any values, will return the margins for the current page.
 *
 * @cat     Document
 * @subcat  Canvas
 * @method  margins
 *
 * @param   {Number} [top] Top margin or all if only one.
 * @param   {Number} [right] Right margin.
 * @param   {Number} [bottom] Bottom margin. Longer descriptions will of course
 *          wrap into the next line again. This happens quite a bit in the
 *          basil.js code base. 
 * @param   {Number} [left] Left margin.
 * @param   {Number} [pageNumber] Sets margins to selected page, currentPage()
 *          if left blank.
 * @return  {Object} Current page margins with the properties: `top`, `right`,
 *          `bottom`, `left`.
 */
```

B
```js
/**
 * @description Sets the margins of a given page. If 1 value is given, all 4
 * sides are set equally. If 4 values are given, the current page will be
 * adjusted. Adding a 5th value will set the margin of a given page. Calling the
 * function without any values, will return the margins for the current page.
 *
 * @cat     Document
 * @subcat  Canvas
 * @method  margins
 *
 * @param   {Number} [top] Top margin or all if only one.
 * @param   {Number} [right] Right margin.
 * @param   {Number} [bottom] Bottom margin. Longer descriptions will of course
 * wrap into the next line again. This happens quite a bit in the basil.js code
 * base. 
 * @param   {Number} [left] Left margin.
 * @param   {Number} [pageNumber] Sets margins to selected page, currentPage()
 * if left blank.
 * @return  {Object} Current page margins with the properties: `top`, `right`,
 * `bottom`, `left`.
 */
```

trych:
Wait, what?

[Here](https://github.com/basiljs/basil.js/issues/266#issuecomment-376915880) you specifically stated you were undecided about having description text all in one line because you said, and I quote

> Might reduce readability when user has no line-wrap enabled

So... you changed your mind now?

trych:
I mean have it all in one line I find easier to maintain, but it renders awefully on GitHub. And others (p5.js, Processing) also seem to avoid this ...

fabianmoronzirfas:
Well. Yes. Seeing the text broken up in favour of line length makes me think that this does not make sense. But we don't need to fight about it. I just think having to indent the text of a description will not happen the time. We will have comments in different styles in the future.

trych:
> We will have comments in different styles in the future.

What do you mean by that? That jsdocs/documentation.js is replaced at some point? Or that the 80-character limit is slowly fading away?

trych:
Yes, but then it's easy to catch in a PR review. And also, someone will forget to style the comments either way, no matter how we decide. But okay, I also don't want to fight, yet I still want have a decision on what we should do.

I myself am a bit torn, I now put hours of work into getting it to look like this in the whole thing, but I can also understand the appeal of having no line breaks at all.

Any opinions from the others? @basiljs

Issue #264, opened 2018-03-23
Function Categories in Reference


trych:
As we discussed, I had a look at the categorizing of basil functions and re-ordered and re-structured them to better reflect the categories they actually belong in and to be better aligned with the categories of Processing and P5.js

One note on constants: I had a look at Processing's handling of constants and I quite like it, I think we should adopt it: They make a difference between constants that are only supposed to be used as function arguments (`CENTER`, `CORNER` etc.), which do *not* get listed in the reference. Then there are those constants that are not necessarily used as function parameters (`PI`, `TAU` etc.) and those are the only ones that get listed in the reference.

As I said, I think we should do the same, as we have a lot of constants by now and many of them cannot be put into specific categories. That means the only constants we *do* list in the reference would be: `VERSION`, `SCRIPTNAME`, `LOREM` and the mathematical ones (`PI`, `TAU` etc.)

Without further ado, here is the new list:

---
## Structure
-   draw()
-   loop()
-   mode()
-   noLoop()
-   setup()

## Environment
-   delay()
-   frameRate()
-   height
-   inspect()
-   printInfo()
-   projectFolder()
-   prop()
-   size()
-   width

  #### Constants

- SCRIPTNAME
- VERSION

#### Collections
- forEach()

#### Conversion
- binary()
- hex()
- unbinary()
- unhex()

#### CSV
- CSV.decode()
- CSV.delimiter()
- CSV.encode()

#### HashList
- HashList()
- HashList.clear()
- HashList.get()
- HashList.getKeys()
- HashList.getKeysByValues()
- HashList.getSortedKeys()
- HashList.getValues()
- HashList.hasKey()
- HashList.hasValue()
- HashList.remove()
- HashList.set()

#### JSON
- JSON.decode()
- JSON.encode()

#### String Functions
- endsWith()
- isURL()
- join()
- nf()
- nfc()
- nfp()
- nfs()
- split()
- splitTokens()
- startsWith()
- trim()
- trimWord()

#### Type-Check
- isArray()
- isInteger()
- isNumber()
- isString()
- isText()

#### Attributes
- ellipseMode()
- rectMode()
- strokeWeight()

#### Primitives
- arc()
- ellipse()
- line()
- rect()

#### Vertex
- addPath()
- beginShape()
- endShape()
- vertex()

## Input

- shellExecute()

#### Files
- download()
- file()
- files()
- folder()
- loadString()
- loadStrings()
- selectFile()
- selectFiles()
- selectFolder()

[...]

PR #277, opened 2018-04-23
Function Categories

trych:
As we discussed in our March Skype call, this updates many function categories, as described in #264. At the same time functions are moved to the files corresponding to their category and sub-category (which is why we have such a large diff, no way around this).

Some notes:
- Each function is now in the file of its main category.
- All functions are grouped by their sub-category (if any) and then sorted alphabetically.
- Categories and sub-categories now align much better with Processing/p5.js where applicable, or a moved to a more suiting basil.js category such as `Document/Page Items`
- Private functions are moved to the end of the category they belong to and then sorted alphabetically.
- Private functions that do belong to several categories are in `core.js`
- Files added: `document.js`, `input.js`, `output.js`.
- Files removed: `public-vars.js`, `global-functions.js` (as they were empty after the sorting)
- `transformation.js` was renamed to `transform.js` to reflect the `Transform` category.
- Section headers were added to each category and sub-category for easy orientation in the bundled `basil.js` file.

So, everything is much cleaner now, and more importantly, it is both easier to find functions as they are consistently placed in their category and it is now easy to know where to put new functions in the future, as they have a dedicated place now. No more guessing and searching around.

This does not affect functionality in any way, to be sure, I ran the test suite and – for good measure – ran all the example scripts. Everything works just like before.

Will go ahead and merge this.


fabianmoronzirfas:
👏 🚀

b-g:
Wow! 🙏

PR #286, opened 2018-10-14
Fixing page functions to return page objects and to jump to new page

trych:
There was an error introduced by my recent page and master page additions/fixes:
`addPage()` did not jump to the newly added page anymore. This fixes it and also fixes return values for `nextPage()` and `previousPage()`.

Will merge right away.

### DIFF
--- a/basil.js
+++ b/basil.js
@@ -2956,7 +2956,7 @@ pub.addPage = function(location) {
     location = pub.AT_END;
   }
 
-  return currentDoc().pages.add(location, pub.page());
+  return getAndUpdatePage(currentDoc().pages.add(location, pub.page()), "addPage");
 };
 
 /**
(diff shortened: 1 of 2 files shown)


ffd8:
Great, ran into this last week and wondered if I missed the discussion about adding pages without going to them. Thanks

trych:
> Great, ran into this last week
Ah, cool, did you also start using (pre) basil 2.0 in your classes then? I did and it runs really smoothly so far. :)

ffd8:
Yupyup, just started teaching with it two weeks ago (made a custom dev build, including my scrappy 'points' functions which still needs some love...) – but it's been a huge joy to use! Suuuuper smooth and fluid– many thanks again for all the great work, @fabianmoronzirfas too! Only hitting minor hiccups (addPage/background), that will be good to get out of the way this semester in our classes before launching? Besides some nostalgic loss.. not missing the `b.` at all! 

Probably not the right place to share it, but here's a new cheat-sheet I made: 
[basiljs_cheatsheet_v2.pdf](https://github.com/basiljs/basil.js/files/2480578/basiljs_cheatsheet_v2.pdf)

PR #287, opened 2018-10-14
Image text placing

trych:
As discussed in #145, this allows placing images in text frames and text into shapes or onto graphic lines (as text paths).

Syntax:
```js
text(txt, obj);
image(imgFile, obj);
```

Added some examples to the `text()` function.

Will merge right away.


trych:
@ffd8 Could you post a code sample of what does not work? I don't quite understand what you mean.

PR #288, opened 2018-10-15
adding background() function, like Processing + p5js

ffd8:
adding compatibility between Processing, P5.js, basil.js.
Based on fill() – so it can accept the same things.


trych:
Hey, cool, I was about to work on that one, too! But no worries, hadn't started yet. :)

I think it would actually be cool, if this sets the background of the entire document, which is something I would solve with master pages, that is place a bg-colored rectangle on each master spread, also I would stretch the background from bleed to bleed to ensure that the entire background is colored.

What do you think?

Also, the description needs to be adapted for `background()`. :)

ffd8:
Aha- whooops, went too fast (just before class), indeed will fix the description.

Haha, don't mind where it goes.. yeah, took a while to decide on document.. i figured things like a guideX/guideY are similar to a background... but happy to move it. (almost fit in 'shapes'.. but sure color fits).

As for the idea about master pages, every page, bleeds, etc.. I would just keep it simple for now and see how people use it. I imagine scenarios where one is making a book, adding pages and wants to have different backgrounds on each page. If they set the canvasMode to bleed, they get bleed-backgrounds, if not, then not.. because it's also nice to work in `canvasMode(MARGINS)` and get a background as expected within that limited size.

trych:
Ok, agree on the size of the background, but I still think it should be placed in the entire document.

Especially when we want to make sure, we can transfer p5.js and processing scripts. There the background command is often in the beginning (setup). When I then add new pages to the document those would not have the background color. If I add it to the master pages, all my new pages will have the background color.

I think the background command is meant to quickly set the background for the *entire* document, so that's what I expect it to do.

If I want different colored pages, I need different commands anyways and can just place rectangles.

fabianmoronzirfas:
Can there be a flag to add it to master for the whole doc or to the page only?

trych:
I *really* think this function should apply to the entire document.

As it is currently implemented we could just easily use a `rect()` command to do the same thing, no need for a new function. However, if we place the background in the entire document, including new pages, that would add some real value and make things easier for the user.

I also think this is closest to the idea of the `background()` function in Processing in p5.js (I know there are no pages, but what I mean is that `background()` simply adds the background and you don't have to think about it anymore, that's for me the equivalent of adding it to the entire document).

trych:
I *really* think this function should apply to the entire document.

As it is currently implemented we could just easily use a `rect()` command to do the same thing, no need for a new function. However, if we place the background in the entire document, including new pages, that would add some real value and make things easier for the user.

I also think this is closest to the idea of the `background()` function in Processing in p5.js (I know there are no pages, but what I mean is that `background()` simply adds the background and you don't have to think about it anymore, that's for me the equivalent of adding it to the entire document).

trych:
Also, with spread-based documents the default behavior as it is now the code

```
page(2);
background(255, 0, 0);
```
would result in something like this:

20181022-120639_screenshot_indesign

Which I don't find matches the idea of a `background()` function very well (and again, one could just easily use `rect()` to get there, so I don't see why we would need a background function with this behavior):

trych:
Also, with spread-based documents the default behavior as it is now the code

```
page(2);
background(255, 0, 0);
```
would result in something like this:

20181022-120639_screenshot_indesign

Which I don't find matches the idea of a `background()` function very well (and again, one could just easily use `rect()` to get there, so I don't see why we would need a background function with this behavior):

trych:
Also, with spread-based documents the default behavior as it is now the code

```
page(2);
background(255, 0, 0);
```
would result in something like this:

20181022-120639_screenshot_indesign

Which I don't find matches the idea of a `background()` function very well (and again, one could just easily use `rect()` to get there, so I don't see why we would need a background function with this behavior):

ffd8:
Hmm I'm guessing we probably won't convince each other either way, haha. Really hard to compare with Processing/P5, other than it _fills the visible canvas with a background_. Perhaps analogous to the pages, is if one works with PGraphic layers? I wouldn't want the background automatically landing on those.. there I specifically draw a background to each buffer if desired. Additionally, per render of `draw()` I clarify if a background should be drawn.. rather than saying it once and getting it forever after. 

Of course the `background()` function should act exactly that of colorful `rect()` – sitting on top whenever it's called (covering up anything under it) – but in this context of having layers, I thought it's best to arrange to the very bottom layer. And of course, in any of these instances, one can use two lines of code to replace the function.. but it's being created for ease, so it should stay easy.  

1 function of only 1-4 params
```
background(255, 0, 0);
```
versus 2 functions, each needing 1-4 params + variables. 
(until now, been doing this often)
```
fill(255, 0, 0);
rect(0, 0, width, height);
```

Sorry, but the example with spreads, reinforces the notion that a background should only fill as much as the `canvasMode()` is set to.. if set it to use `FACING_PAGES`, then the background will fill both.. otherwise, just the single page environment they're working in.

I imagine each new added page to be the next rendering of the `draw()` function, with a clean slate.. but as mention above, gets really weird when using multiple master pages. Therefore, I think we could/should make a nice tutorial about editing a master page before starting if such a background etc is desired. (which wasn't previously so easy until this new great 2.0 function for `masterPage()`)

ffd8:
Trying to think of a compromise.. was thinking maybe it should work with a function that allows one to apply an individual thing to the masterpage, or perhaps, could add another param to the background() for which masterpage.. just quickly turns into strange number of params:
`background(150); // grayscale BG on current page`
`background(150, "A-Master"); // grayscale BG on master page`
`background(0, 255, 0, "A-Master"); // RGB BG on master page `
`background(0, 100, 0, 0, "A-Master"); // CMYK BG on master page = ugh.. `

but not sooo bad:
`background(gradient(c1, c2)); // gradient BG on current page`
`background(gradient(c1, c2), "A-Master"); // gradient BG on master page`

Will require quite a few conditionals... but should work.
Thoughts?

trych:
Hi @ffd8,
I have a crazy deadline coming up on Monday, so I can't really spend to much time thinking about this before then. However, judging the compromise solution quickly, I would say to keep it simpler.

My suggestion would be to allow a document object as an optional parameter and nothing else.
So you could do

```
background(255, 0, 0);
```

to give a background to the current page (or spread, if you are set to facing pages).
Or you could do

`background(255, 0, 0, doc());`

to give a background to the entire document as I have described above (i.e. *all* pages, all masters, up to the bleed, so that really the *entire* document has a red backdrop).
That would also keep the conditionals to an acceptable level, as we could simply check, if the last parameter is a document.

And if I have the scenario you described and wanted to give only a background to a specific master, you could just do this:

```
masterPage("A"); // jump to master page A
background(255, 0, 0);
```

(Also on a site note, does anybody know why the `masterPage` function does not show up in the [WIP basil reference page](https://basiljs2.netlify.com/reference/)?)

fabianmoronzirfas:
> (Also on a site note, does anybody know why the masterPage function does not show up in the WIP basil reference page?) . 

Update to latest version. https://basiljs2.netlify.com/

The site does not update (yet) on changes done to basil.js develop branch. Only on push to the develop branch of the basiljs.github.io repo. 
I hope that [github/actions](https://github.com/features/actions) will solve this for us.

trych:
@fabianmoronzirfas and @b-g, could you guys maybe state your opinion on how this function should work? I am against a flag to allow for both options, because then this function becomes really messy.

So, the two options are basically:

A) Add a background color to the current page only.
B) Add a background color to the entire document by adding it on all master pages.

For detailed pros and cons see above. Since Ted and I have differing preferences, I would say you guys should decide, so we can move ahead with this.

fabianmoronzirfas:
Read the thread. I'm for solution A even though it could be achieved with a rect 

> A) Add a background color to the current page only.

trych:
Okay, A) it is then. :)
I'll review the code and then we can go ahead and merge it. (Or if anyone else wants to review it, go ahead)

trych:
Ok, I have made a few changes:

- cleanup (mainly got rid of the else if cascade via the `.apply()` method that I just found out about 😄)
- got rid of the layer clearing, because that would not work if you set backgrounds on several pages, it would clear all backgrounds that you set previously on different pages. Instead I would just leave the shapes where they are, because with the way you proposed the funciton some user could try something like this:
    ```js
    canvasMode(FACING_PAGES);
    background(255, 0, 0);

    canvasMode(MARGIN);
    background(255, 255, 0);
    ```
    in which case the red background that spans the entire spread should not be deleted again.
- made sure the backgrounds are drawn without a stroke

Could someone else have another look at this?

trych:
Ok, this should not be merged yet, as there is still one unsolved problem:

If I run a script this script:

```js
clear(doc());
ellipse(50, 50, 50, 50);
background(255, 0, 0);
```

it works fine. However, when I then make the `basil.js Background` layer the active layer and run the script again, the oval will be placed on the background layer and then the background rectangle will be placed in front of it.

How should we resolve this issue?

ffd8:
@trych Perhaps to solve this, the background layer should be unlocked/locked from start to finish of function? If they try to write to that layer, they'll get an InDesign warning, which is fair enough since they probably didn't realize it was selected.

trych:
> If they try to write to that layer, they'll get an InDesign warning, which is fair enough since they probably didn't realize it was selected.

That gives me the idea that basil.js could implement some mechanism to automatically avoid drawing shapes on a locked layer and automatically move to the next higher layer when drawing a shape, while printing a friendly warning to the console. That would be nice. Would probably require some global `currentLayerIsLocked` flag or something.

However, that is a topic for a different feature.
I think your suggestion to unlock the layer in the beginning and lock it after running the command is fine.

(I am still not a big fan of the idea to draw the background on only one page. Seems like an awefully complicated function just to place a rectangle in the back. But I guess I was overruled …)

ffd8:
`background()` should be fixed now, but it exposed an interesting issue with our `clear()`. When a `layer()` is sent, it should probably only clear that page layer's contents, right? Otherwise one should just `remove()` the layer altogether. This explains why `nameOnPage()` was always working best for that given page then calling by `layer()` to isolate items on that particular page... While I implemented the following for-loop for this `background()` function, it's probably best to modify our `clear()` function with the following:

```
if (container instanceof === Layer){
  var containerItems = container.pageItems;
  for(var i=0; i < containerItems.length; i++){
    var curItem = containerItems[i];
    if(curItem.parentPage.name === currentPage().name){
      curItem.remove();
    }
  }
}
```

trych:
> When a layer() is sent, it should probably only clear that page layer's contents, right?

Why? When I pass a layer I expect it to clear the layer in the entire document. And I don't see any reason to change this. This is actually quite useful if I selectively only need to set and later reset certain parts of a document. I use this a lot.

But having said that, it would be super nice if the clear function would allow to pass some parameters that could be checked.

Something like this:

```js
clear(myPage, {
  itemLayer: "foobar",
  strokeWeight: 3}
);
```
would only clear those page items on the page that are on layer foobar and have a strokeWeight of 3.

basil.js
```diff
@@ -860,6 +860,48 @@ var clearConsole = function() {
...
+ * @param   {Color|Gradient|Swatch|Numbers|String} bgColor Accepts a color/gradient/swatch as string name or variable. Or values: GRAY / R,G,B / C,M,Y,K.
+ * @return  {Rectangle} The newly created background shape.
+ */
+pub.background = function (bgColor) {
+
+  var pLayer = currentLayer();
+  var bgLayerName = "basil.js Background";
+  var bgLayer = pub.layer(bgLayerName);
+
+  var lockedState = bgLayer.locked;
```
trych: We don't need `lockedState` anymore, should be removed.

basil.js
```diff
@@ -860,6 +860,48 @@ var clearConsole = function() {
...
+  bgLayer.locked = false;
+  pub.arrange(bgLayer, pub.BACK);
```
trych: I think it should be only arranged back if it is newly created. When it already exists and there is a layer underneath it, it is probably done so on purpose.

basil.js
```diff
@@ -860,6 +860,48 @@ var clearConsole = function() {
...
+
+  var backgroundShape = pub.rect(0, 0, pub.width, pub.height);
+  backgroundShape.fillColor = getSwatch("background", arguments);
+  backgroundShape.strokeWeight = 0;
+
+  bgLayer.locked = true;
+  if(pLayer.name != bgLayerName){
+    pub.layer(pLayer);
+  }else{
+    pub.layer('Layer 1');
```
trych: Should this not be something like `pub.layer(pLayer)`? `Layer 1` is not even available in many documents or non-English ID installations.

basil.js
```diff
@@ -860,6 +860,48 @@ var clearConsole = function() {
...
+  var bgLayer = pub.layer(bgLayerName);
+
+  var lockedState = bgLayer.locked;
+  bgLayer.locked = false;
+  pub.arrange(bgLayer, pub.BACK);
+  
+  var bgItems = bgLayer.pageItems;
+  for(var i=0; i < bgItems.length; i++){
+    var curItem = bgItems[i];
+    if(curItem.parentPage.name === currentPage().name){
```
trych: I think the check should be rather something like

```js
if(curItem.parentPage === currentPage())
```

(does this work?)
Page name can be ambiguous. E.g. there can be two pages named `1` in the same document.

ffd8:
Thanks for the comments above – took all into account and created a test. 
Regarding the pLayer, see comment above (though it mentions outdated) – but only applies to if background layer had been selected (which is locked) – now it checks if there is more than 1 layer and will select topmost otherwise create an 'output' layer for drawing anything after the background (again, only applies if background had been selected layer at runtime).

basil.js
```diff
@@ -876,14 +876,15 @@ pub.background = function (bgColor) {
   var bgLayerName = "basil.js Background";
   var bgLayer = pub.layer(bgLayerName);
 
-  var lockedState = bgLayer.locked;
   bgLayer.locked = false;
-  pub.arrange(bgLayer, pub.BACK);
+  if (currentDoc().layers[0].name === bgLayerName) {
```
trych: I think either you misunderstood me, or I am not getting what you are trying to achieve here. What I meant was that the bgLayer should be moved to the back only if it was newly created, not necessarily always when it is in front. (it could be in front when it already exists as the top layer).

basil.js
```diff
@@ -893,10 +894,15 @@ pub.background = function (bgColor) {
...
 
   bgLayer.locked = true;
-  if(pLayer.name != bgLayerName){
-    pub.layer(pLayer);
+  if (pLayer.name === bgLayerName) {
+    var docLayers = currentDoc().layers;
+    if (docLayers.length > 1) {
+      pub.layer(docLayers[0]);
+    }else{
+      pub.layer('output');
```
trych: I don't think we should pick an arbitrary name. Better let InDesign create the layer and name it according to the user's locale.
Also, I think it would be more natural not to select the foremost layer, but the layer right before the bgLayer (as this one was selected previously, so the selection does not jump away very far.

trych:
See my additional comments above. I prepared a branch with changes, which I will PR against this branch, so you can see, if the changes make sense.

Also, on a different note, your editor does not seem to trim trailing whitespace on saving. Could you maybe activate this. Else we'll end up with a lot of diffs where just whitespace is trimmed. Thanks.

basil.js
ffd8: Yup that's what I thought was happening here.. only if the layer was JUST created, would it be sent back.. but true.. incase it's for some reason the top layer – then would get unnecessarily forced back. Your approach of checking layer count before making it is better.

basil.js
ffd8: Sounds good – adding a generic layer instead of one called 'output'.

ffd8:
Changes sound good – let's merge them.

Regarding editor and whitespaces – just did a search and modified sublime's prefs to prevent it. Thanks for the heads up.

trych:
Merged at last. Well, that only took us 1 1/2 years. 😆 😉

trych:
Or, to stay in the meme parlance of our times:

20200310-203711_Screenshot_Firefox

Issue #120, opened 2016-11-07
Easier transforming of images


trych:
At the moment to transform an image `b.transformImage(img, x, y, width, height)` has to be used. All the `b.itemX`, `b.itemWidth` and so on commands cannot be used as setters, as only the frame will be transformed, not the image itself.

If I for example want to scale an image in place, I will have to get all four values just to do that: x, y, width, height. Same thing, if I want to move it to another position without scaling, I still have to get width and height first.

To make working with this easier we could either

1. Introduce methods like `b.imageX()`, `b.imageY()` and so on. Or simpler `b.translateImage()` and `b.scaleImage()` etc.

OR

2. (which I think would be better) Modify the behavior of `b.itemX()`, `b.itemWidth()` that they set the `x`, `width` etc. for *both* the image frame and it's contained graphic. In this scenario we would need to introduce a method to get the contained graphic somehow, in case this should be transformed by itself (which is a method I wanted to suggest anyways, as it can be very useful). So something like `b.graphic(myImageFrame)`. This solution would keep everything clean in the sense that `b.itemWidth()` and `b.itemSize()` and so on can be used consistently for any item.

Thoughts?

trych:
Ok, after thinking about this for a while now, version 1 does not make any sense to me anymore. I think version two is much cleaner and the way to go. Thus my suggestions:

1. Make all `b.itemXXX()` functions work for images, meaning treat any graphic as a child of its frame and transform it along.

2. Introduce some method to get the graphic of a frame, so it can be treated separately if the need arises. I would simply suggest `b.graphic(frame)`.

3.) Keep `b.transformImage()` for backwards compatibility reasons only. (Or we decide that the big new version 2.0 can break some compatibility and get rid of it altogether).

b-g:
Hi @trych, many thanks for the thoughts! Here's my feedback:

1. OK + even better + good call!
2. I would rather name it then `b. graphics` (note the 's') and implement it analog to b.selections, b.words, b.stories
3. As we introduce dramatic new things with 2.0, I would strongly vote to get rid of legacy things

trych:
Cool! I'm fine with getting rid of it, as it is rather confusing to have a separate function in the first place.

Issue #72, opened 2016-07-22
Replace b.addToStory() by b.addToText


trych:
Apart from the fact that `b.addToStory()` does not work at the moment ( #71 ) , I think it could also improved structurally. This was mentioned in another issue ( #44 ), so I will mostly copy/paste to this issue here.

I think the method could become more "basil like" beginner friendly, if it was changed in these ways:

**1)** Skip the requirement to have a `story` object instance as input. It is quite difficult to retrieve a specific `story` object within the basil.js universe and I guess many beginners don't know what a `story` actually is. How it could work instead: Allow `textFrames` as input. `textFrames` are a whole lot easier to address in basil.js and a much easier and much less concept for beginners. The new texts could be added at the beginning at the end or somewhere within the `textFrame`. Should text then move to the next linked `textFrame` or into overset text, it will be obvious. Additionally we could allow `paragraphs`, `lines`, `words`, `characters` (and yes, why not, `stories`) as input. This could also be convenient, if you found a `word` already then you could easily just add text to it. So the whole method rather becomes a "`b.addToText()`" (and should be renamed accordingly) and easier to use, I think.

**2)** Instead of an `insertionPoint` as input I would also allow for a simple index. Same as with `stories`, there is no way in basil to retrieve an `insertionPoint` (or is there?) and beginners don't know what an `insertionPoint` is (let alone that it is an object, not a "index like" position). So if we give the option to use a simple number as index, I think it would be also much easier for beginners.

And we could also include the "add in certain (paragraph|character) style" feature that @b-g proposed in #44.

So the specific method would be:

`b.addText({textframe|story|paragraph|line|word|character}, text, [index|insertionPoint|b.AT_BEGINNING|b.AT_END],[characterStyle|paragraphStyle] ) `

If others like the idea, too, the question is, if we should then discontinue the b.addToStory() method. Usually I would be all for staying backwards compatible, but this feature has not been working for almost three years now (does not work in 1.08), so I think we could take it out of the official reference and replace it by b.addToText().

We could still leave a b.addToStory method in the code that throws a basil.js warning: b.addToStory(), this method is deprecated. Please use b.addToText() instead. Or simply let the b.addToStory() method run the b.addToText() method (as it is compatible regarding the inputs).

Please flag as enhancement and feature.

trych:
[...]

PR #289, opened 2018-10-15
Rework of collection functions + graphics() + textStyleRanges()

trych:
This standardizes and cleans up the collection functions (e.g., `words()`, `characters`, `items()` etc.) and adds two new ones: `graphics()` and `textStyleRanges()`.

Changes in detail:
- collection functions now always return collections (or arrays), also when a callback function is used
- if a callback functions is used that returns falls at some point and interrupts the loop, a collection of elements up to that point is returned
- same behaviour goes for `forEach()`
- legal containers were made more consistent between functions and additional possible containers were added for many functions (e.g., words on a layer, words in a group, items in a paragraph etc.)
- docs were improved to be easier to understand (I hope)
- internal cleanup, including the removal of private function `forEachTextCollection()` which was no longer needed
- `graphics()` was added as discussed in #120 to allow changing the graphics within containers, like the graphic in a graphic frame.
- `textStyleRanges()` was added as discussed in #72 to allow looping over, well, text style ranges. :)
- `stories()` now allows for text frames or text objects as containers to get their parent story, as discussed in #72.

@basiljs I will leave this up for review for a day or so, if there are no vetoes, I will go ahead and merge.

### DIFF
--- a/basil.js
+++ b/basil.js
@@ -1524,10 +1524,13 @@ forEach = function(collection, cb) {
     }
 
     if(cb(collection[i], i) === false) {
-      return false;
+      if(collection.hasOwnProperty("everyItem")) {
+        return collection.everyItem().getElements().slice(0, i);
+      }
+      return collection.slice(0, i);
     }
   }
-  return true;
+  return collection;
 };
 
 // ----------------------------------------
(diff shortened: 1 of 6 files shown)


fabianmoronzirfas:
Cool thing. What about some test coverage for the new functions?

trych:
Added some testing.
Feel free to add to this testing (or any testing for this matter) whenever you feel, there are specific features that should be tested.

Will merge now.

PR #294, opened 2018-10-18
point(), quad() and triangle()

trych:
In the effort to make basil.js more similar to p5.js and Processing, this adds the shape functions `point()`, `triangle()` and `quad()`, which should be all really self explanatory.

Additionally this allows for a three parameter input for `ellipse()` if the user wants to create a circle.

Will merge this right away.

### DIFF
--- a/basil.js
+++ b/basil.js
@@ -6839,12 +6839,13 @@ pub.quad = function(x1, y1, x2, y2, x3, y3, x4, y4) {
     error("quad(), not enough parameters to draw a quad! Use: x1, y1, x2, y2, x3, y3, x4, y4");
   }
 
-  beginShape();
-    vertex(x1, y1);
-    vertex(x2, y2);
-    vertex(x3, y3);
-    vertex(x4, y4);
-  return endShape(CLOSE);
+  var q = addShape([[x1, y1], [x2, y2], [x3, y3], [x4, y4]]);
+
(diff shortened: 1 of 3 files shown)


fabianmoronzirfas:
I think something like this won't work. 

```js
function setup(){
function pseudoPointClass (x, y){

this.display = function(){
  this.point = point(x, y);
  this.point.x = 10; // this is pseudo code
}
}
this.update(){
// something else here
}
var points = [];
for(var i = 0; i < 10; i++){
points.push(new pseudoPointClass(10, i*10));
}
// do some stuff
for(var i = 0; i < points.length; i ++){
stroke(i*10);
points[i].display();
points[i].update();
}
}
```

trych:
@fabianmoronzirfas @ffd8 I am getting notification emails about posts from both of you, however I don't see these posts here on the site. Is this stuff that you posted but then deleted again? Or is there something wrong with my Github?

trych:
@fabianmoronzirfas @ffd8 I am getting notification emails about posts from both of you, however I don't see these posts here on the site. Is this stuff that you posted but then deleted again? Or is there something wrong with my Github?

ffd8:
there were weird problems commenting last night, as github had some [dataserver issues](https://status.github.com/messages). I left you a comment in the pull request of your `image()` update, that placing in textframe doesn't seem to be working and a comment on my PR for `background()`

Issue #238, opened 2017-12-26
b.selection() as setter


trych:
I think `b.selection()` should also allow to set the selection.

Why on earth would I need that, you ask? 😉
It would be very useful for PNG export with selected elements only.

fabianmoronzirfas:
Yes. Sounds good

trych:
Implemented in #297 . Closing.

PR #297, opened 2018-11-12
selection() as setter

trych:
Allows to use `selection()` as setter as explained in #238 .
Will merge right away.

### DIFF
--- a/basil.js
+++ b/basil.js
@@ -3641,15 +3641,21 @@ pub.objectStyle = function(itemOrName, props) {
 };
 
 /**
- * @description Returns the first currently selected object. Use this if you know you only have one selected item and don't want to deal with an array.
+ * @description If no argument is given, returns the first currently selected object. If a page item is given as argument, the page item will be selected.
  *
  * @cat     Document
  * @subcat  Page Items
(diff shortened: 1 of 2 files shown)


ffd8:
Eek, I somehow hadn't used `selection()` since this change. But it killed the previous function of 'selection()' to grab whatever was selected within InDesign interface?? Make a textFrame or polygon within InDesign GUI, select it, then run:

```
var sel = selection();
inspect(sel);
```

Returns 'undefined is not an object'. 
It shouldn't require param, so that it can still be used to grab whatever pageItem was clicked on (we've used it for making scripts that you run from within InDesign, selecting something then presto chango to it).

trych:
Yup. I did not want to change this behaviour of course, it was just a bug in the implementation. Fixed in the current develop branch (see #330).

ffd8:
Thanks again for fixing the issue. 
Still curious how you're using it as described in #238 ..

trych:
You mean how to use it as a setter?
You just give the object to be selected as an argument and then you could – for example – export a png and in the export setting set it to selection only, so only the selected part will be exported.

Issue #295, opened 2018-11-12
measure()


trych:
After I have been using the new `transform()` function for a while now, I came to the conclusion that it feels quite awkward to use it for measuring something. Example:

```js
transform(myPageItem, "rotation");
```

Although we decided initially to create transform as a setter and getter, the term does not really work for a getter. If I am just using it to get the rotation of an item, I am not transforming anything, hence the term feels wrong. My suggestions therefore is to introduce a function named `measure()` that allows me to do exactly that: measure transform properties of an object.

It could simply be a wrapper of `transform()` that allows getting values, but I think, if we have this `measure()` function anyways, we could make it even more useful. I think it could

- replace `bounds()` and therefore also measure typographic features (in case my item is text).
- unlike `bounds()` it could measure specific properties only. Example:

    ```js
    measure(myWord, "baseline");
    ```
    could measure the baseline of each word only and would therefor be much faster than `bounds()`  measuring and calculating all the properties.
- it could be used to measure page and spread bounds as well

I think the syntax `measure(myItem, "property");` is a bit easier and more basil-like than `bounds(myWord).property`. Since we can make breaking changes before 2.0, now would be the best time to make a change like this.

@basiljs What do you guys think?

ffd8:
I think it's an issue of semantics for 'transform', which does imply that one is actually changing something. It's normal with the `typo()` function among others to use it as getter/setter.. Additionally `property()` isn't working as a getter, which it could/should? Maybe `transform()` is still the wrong word for what it's doing? Should it be merged with and extending `property()`? (haha which I'd also shorten to just `prop()` to match `typo()`). 

Nevertheless `bounds()` is really useful, since I'm/we're typically using the 4-5 properties returned (left, top, width, height, baseline):

```
var bnds = bounds(obj);
rect(bnds.left, bnds.top, bnds.width, bnds.height);
```
which has been fine for students... but true, there aren't so many getters that return an object.

trych:
Yes, exactly, it is a semantic issue. That's why I would like to add `measure()` so skip the awkward `transform(item, "property")` for measuring.

The bounds thing could simply be solved by offering the possibility to do `measure(item, "bounds")` and then this could give us the object that we know from bounds. I thinkt it would be nicer to have all this in one function called `measure()`.

Also, the `bounds()` function is quite broken, so we would need to fix this anyways, so we might just implement it in `measure()` instead.

trych:
@fabianmoronzirfas @b-g
[...]

PR #298, opened 2018-11-12
Dev fixbounds

ffd8:
Fixed a few issues in `bounds()` when using `canvasMode(MARGIN)` or `canvasMode(FACING_MARGINS)`. `BLEED`, `FACING_BLEEDS`, `PAGE`, `FACING_PAGES` are all working fine too.

Only thing that still remains, is when getting the `bounds(pages())` in one of the `FACING_` modes, which was already an issue and can be solved later.

Here's a test script using `bounds()` of a few items:

```
// @includepath ~/Documents/;%USERPROFILE%Documents; 
// @include basiljs/basil.js; 
 
function draw() { 
  var output = layer("output");
  clear(output);
  
  canvasMode(MARGIN);
  //canvasMode(FACING_MARGINS);

  // first rect
  var r = rect(0, 0, width, height);
  var bnds = bounds(r);
  
  // 2nd rect overlay
  fill(255, 0, 0);
  var r2 = rect(bnds.left, bnds.top, bnds.width, bnds.height);
 
  // text, with bounds of letter
  fill(0, 255, 0);
  var tf = text("hello", 0, 0, width, height);
  typo(tf, "pointSize", 150);
  characters(tf, function(obj){
    if(obj.contents.toLowerCase() == 'h'){
      var bnds = bounds(obj);
      rectMode(CORNERS);
      fill(0, 0, 255);
      var r= rect(bnds.left, bnds.xHeight, bnds.right, bnds.baseline);
      referencePoint(CENTER);
      transform(r, "rotation", 45);
    }
  });

  // issue with bounds(page()) remains in FACING mode
  var bnd = bounds(page());
  fill(gradient(color(50), color(200)));
  var bg = rect(bnd.left, bnd.top, bnd.width, bnd.height);
  arrange(bg, BACK);
} 
```

### DIFF
--- a/basil.js
+++ b/basil.js
@@ -3390,8 +3390,8 @@ pub.bounds = function (obj) {
       y1 = bounds[0] - currOriginY;
       x2 = bounds[3] - currOriginX;
       y2 = bounds[2] - currOriginY;
-      w = x2 - x1;
-      h = y2 - y1;
+      w = x2 - x1 - currOriginX;
+      h = y2 - y1 - currOriginX;
       return {"width": w, "height": h, "left": x1, "right": x2, "top": y1, "bottom": y2};
     }
     // no idea what that might be, give up
--- a/basil.js
+++ b/basil.js
@@ -3365,20 +3365,20 @@ pub.bounds = function (obj) {
 
     return {"width": w,
             "height": h,
-            "left": x1 + currOriginX,
-            "right": x2 + currOriginX,
-            "top": y1 + currOriginY,
-            "bottom": y2 + currOriginY,
-            "baseline": baseline + currOriginY,
-            "xHeight": xHeight + currOriginY};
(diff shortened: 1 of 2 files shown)


trych:
@ffd8 There is a large list of issues with `bounds()` still, as you can see in #166. It basically needs a proper rewrite. I started a branch on this once and it soon turned into a huge task. That's why I would vote to not implement a fix hastily now that leaves many of the problems in place, but to either rewrite the function to adress these issues, or – and that's what I am voting for – to replace the bounds function by `measure()`, as I suggested in #295 .

The bounds functionality itself could be kept by implementing something like `measure(item, "bounds")`. Either way. All measuring should be handled by the inner functionality of `transform()` as this is currently the only function that handles measurements correctly, even with applied matrix transformations.

I would vote to close this pull request and not merge it.

PR #312, opened 2018-12-10
pageCount fix

ffd8:
It had caused error if changing `pageCount()` to a number less than document had. Useful for undo'ing a script that regularly uses `addPage()`. Switches to last available page to avoid the error.


ffd8:
Ahhh had wrong branch selected... possible to really trash this or simply closing is enough?

PR #313, opened 2018-12-10
Fixing pageCount

ffd8:
Caused error if changing `pageCount()` to anything less than current document had. Useful for undo'ing a script that regularly uses `addPage()`. Now will auto switch to last available page to avoid crash.

### DIFF
--- a/basil.js
+++ b/basil.js
@@ -3150,8 +3150,12 @@ pub.page = function(page) {
  */
 pub.pageCount = function(pageCount) {
   if(arguments.length) {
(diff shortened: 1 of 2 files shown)


trych:
Hmm, upon first glance, I would only switch to last page in case the current page (before `pageCount()`) is actually removed. If I am on the first page and reduce my document from 30 to 5 pages, I want to stay on the first page and not jump to page 5. ;)

ffd8:
You're right – changing.

PR #314, opened 2018-12-11
Develop pagecount fix

ffd8:
Updated to leave the user on the page they were on if within the new `pageCount()`.


trych:
This looks good and fixes the bug I had. :) Thanks! Will merge.

PR #318, opened 2018-12-27
removing blockcomments from constants

fabianmoronzirfas:
Adds the `@subcat Constants` to all constants (duuh) Should we add the `@cat global` as well?


fabianmoronzirfas:
fixes #318

trych:
What's with this huge diff in `doc/api.json`? Is it supposed to be such a large change, if only the constants are removed?

fabianmoronzirfas:
Actually we could remove the doc/api.json all the generating of docs happens in the site repo

PR #330, opened 2019-03-05
selection() works again with no arguments

trych:
Quick fix for `selection()` to work again with no arguments. Will merge right away.


ffd8:
Thanks for the quick fix!

PR #342, opened 2019-12-16
Add a Code of Conduct

fabianmoronzirfas:
I think we should have on (also if we want to keep on using Netlify we need one)

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,76 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
+contributors and maintainers pledge to making participation in our project and
+our community a harassment-free experience for everyone, regardless of age, body
+size, disability, ethnicity, sex characteristics, gender identity and expression,
+level of experience, education, socio-economic status, nationality, personal
+appearance, race, religion, or sexual identity and orientation.
+
+## Our Standards
+
+Examples of behavior that contributes to creating a positive environment
+include:
+
+* Using welcoming and inclusive language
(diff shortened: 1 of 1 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,76 @@
...
+when an individual is representing the project or its community. Examples of
+representing a project or community include using an official project e-mail
+address, posting via an official social media account, or acting as an appointed
+representative at an online or offline event. Representation of a project may be
+further defined and clarified by project maintainers.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported by contacting the project team at info@technologiestiftung-berlin.de. All
```
b-g: Wrong email address :)

b-g:
@fabianmoronzirfas Many thanks for this! Just found a copy & paste error. See comment. Works for me.

CODE_OF_CONDUCT.md
fabianmoronzirfas: Fixed.

fabianmoronzirfas:
@ffd8 @trych are you okay with this? We need to move forward so we don't lose the OSS Team rights on netlify.

fabianmoronzirfas:
To make this more prominent we could also create a repo for the CoC and just link there in each repo from a document like this one.

trych:
> @ffd8 @trych are you okay with this?

Yes, sure.

ffd8:
Sounds good, though I think it's fine to just have in the main basil.js repo (that's how p5.js is doing it, not for each project, web-editor or website)?

fabianmoronzirfas:
That’s why I thought having a dedicated repo would be good

PR #344, opened 2020-03-03
adding couple week/day functions

ffd8:
- added `week()` for returning value of week throughout year
- added `weekdayCount()` for returning value of weekday
(ran into the need for these during a recent workshop)

### DIFF
--- a/basil.js
+++ b/basil.js
@@ -5643,36 +5643,17 @@ pub.week = function() {
 };
 
 /**
- * @summary Returns the current day of the week.
- * @description The `weekday()` function returns the current weekday as a string from `Sunday`, `Monday`, `Tuesday` ...
+ * @summary Returns the current day number of the week.
+ * @description The `weekday()` function returns the current weekday as a value from `0` - `6` (starting Sunday).
  *
  * @cat     Input
  * @subcat  Time & Date
  * @method  weekday
  *
- * @return  {String} The current weekday name.
- */
-pub.weekday = function() {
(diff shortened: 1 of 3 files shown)


trych (changes requested):
Looks good overall, but regarding the weekday functionality:

I would prefer to just have a function to return the current weekday as a number, name that one `weekday()` and in turn get rid of the current `weekday()` function that returns the English weekday names. So basically, keep your two new functions and kick out the old one.

If we have a function to return the weekdays as numbers, then it's easy enough to write a quick array with names in whatever language you prefer which is much more flexible and much more in the spirit of Processing/p5/basil to be open to all (and not just English users).

What do you think?

Also, please add a quick note on the changes to the changelog.

ffd8:
I agree the printed days of the week probably isn't necessary, especially since it's only English. But only returning numbers becomes a little strange... since JS for the day of the week returns 0 (Sunday) to 6 (Saturday)... based on ISO, should refer to numbers as 1 (Monday) and 7 (Sunday).. ha. Maybe it's best to leave it 0-6 (so it's cycles nicely through someone's custom array.. and if one wants the number like ISO, do some math to the output (like currently implemented)?

trych:
> Maybe it's best to leave it 0-6

Yeah, even I as a German agree. I mean it's easy enough to just move my `"Sonntag"` string to the beginning of my array once I notice my days are off.

ffd8:
Changes made - should be good.

trych (approved):

trych:
Looks good, thanks!

PR #345, opened 2020-03-03
pass array into random

ffd8:
Love this feature of p5.js- pass an array into `random()` and returns a random item from it = so useful when crunching data or setting up simple generative design choices.


trych:
Good idea, was always missing this as well.
Made some minor adjustments (we should try to rely on our own type-checking consistently) and added to the changelog.

Eventually I think we should enable this for all InDesign collections as well, but no need to include that right now.

If you're fine with the changes, please merge!

trych (approved):

ffd8:
Aha - haha thanks, forgot about change log.
Regarding any InDesign collections, yeah that would be a nice next step.

PR #346, opened 2020-03-03
getURL updates - params + Windows support

ffd8:
- added support for CURL optional params (mac only) when using `loadString()` / `loadStrings()` / `loadJSON()`, since some API's (reddit) require sending a `--user-agent` to make a unique request.
- added support for Windows to also load URLs in the above functions! This had been a sore point for PC users in classes/workshops – but tested this script a few times and seems to work fine.

### DIFF
--- a/basil.js
+++ b/basil.js
@@ -5351,14 +5351,15 @@ pub.folder = function(folderPath) {
  * @method  loadJSON
  *
  * @param   {String|File} file The JSON file name in the document's data directory, an absolute path to a JSON file, a File instance or an URL.
+ * @param   {String} [curlOptions] Optional Mac-only parameters when URL is used, ie. `"--user-agent 'browser agent'"`
  * @return  {Object} The resulting data object.
  */
-pub.loadJSON = function(file) {
+pub.loadJSON = function(file, curlOptions) {
 
   var jsonString;
 
   if (isURL(file)) {
-    jsonString = getURL(file);
+    jsonString = getURL(file, curlOptions);
   } else {
     var inputFile = initDataFile(file),
       data = null;
@@ -5379,11 +5380,12 @@ pub.loadJSON = function(file) {
  * @method  loadString
  *
  * @param   {String|File} file The text file name in the document's data directory or a File instance or an URL
+ * @param   {String} [curlOptions] Optional Mac-only parameters when URL is used, ie. `"--user-agent 'browser agent'"`
  * @return  {String} String file or URL content.
  */
-pub.loadString = function(file) {
+pub.loadString = function(file, curlOptions) {
   if (isURL(file)) {
-    return getURL(file);
+    return getURL(file, curlOptions);
   } else {
     var inputFile = initDataFile(file),
       data = null;
@@ -5403,11 +5405,12 @@ pub.loadString = function(file) {
  * @method  loadStrings
  *
  * @param   {String|File} file The text file name in the document's data directory or a file instance or an URL
+ * @param   {String} [curlOptions] Optional Mac-only parameters when URL is used, ie. `"--user-agent 'browser agent'"`
  * @return  {Array} Array of the individual lines in the given file or URL
  */
-pub.loadStrings = function(file) {
+pub.loadStrings = function(file, curlOptions) {
   if (isURL(file)) {
-    var result = getURL(file);
+    var result = getURL(file, curlOptions);
     return result.match(/[^\r\n]+/g);
   } else {
     var inputFile = initDataFile(file),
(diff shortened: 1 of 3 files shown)


ffd8:
I'll try to find a student on Windows to test this.. but anyone on Windows to give it a try?
`println(loadString('https://official-joke-api.appspot.com/jokes/programming/random'));`

trych (changes requested):
I think this curl functionality should be moved to a plugin, especially when it is Mac only, I also think it is too specific to be included in such a function. We can always use the `shellExecute()` function, if we need to get really specific with some shell commands.

With the download function on windows:
While I am happy that there is something working on Windows, I would much prefer to have the `shellExecute()` function itself work on Windows and then have the download functions rely on that internally instead of having a really specific download-only windows shell functionality within the download function.

I remember I started working on this a while back, but never finished it. Maybe what you have now in the download function could be used as a starting point to get the shell function in Windows working?

ffd8:
Just a reminder– this isn't about the downloading of things (images etc) – just the grabbing of strings from remote location. Our current `getURL()` has always been using curl (and was always mac only) – this change opens up that functionality to Windows. The second modification is for pretty specific curl options like --user-agent. I don't think the curl should be a plugin – since it should be out of the box that one can pass a URL to `loadString()` etc.

While the `shellExecute()` could use some universal OS love, this should patch a problem that's existed for years.

trych:
> Just a reminder– this isn't about the downloading of things (images etc) – just the grabbing of strings from remote location. Our current `getURL()` has always been using curl (and was always mac only) – this change opens up that functionality to Windows.

trych:
Sorry, clicked the wrong button.

trych:
> Just a reminder– this isn't about the downloading of things (images etc) – just the grabbing of strings from remote location. Our current `getURL()` has always been using curl (and was always mac only) – this change opens up that functionality to Windows.

Hm. Okay. While I still think this should internally use `shellExecute()`, there is no harm including that part for now to unlock the downloading for Windows users.

> The second modification is for pretty specific curl options like --user-agent.

Exactly, since these modifications are pretty specific, I would not want to include them in the general download functions. Those who need these specific options should be advanced enough to easily use the `shellExecute()` function and pass those parameters there.
And I am 100% against adding yet another Mac-only feature, as I was – quite the opposite – hoping to close this Mac-Windows gap in basil altogether soon.

If you insist of having some Mac-only functionality added somewhere, then it's not a problem to add it as a plugin.

ffd8:
FYI – downloading (images) remains non-existant for Windows (can't work on that, as I'm Mac based). Nevertheless, this is just about scraping/pinging information from the internet – APIs, etc. and updating our existing functions for that. 

It's now changed to allow a custom user-agent for both Mac + Windows(!). Seems special, but I was surprised that it was necessary during a workshop recently when using Reddit's GET API (and student was a complete beginner – wouldn't be doing any `shellExecute()` themselves).

Reddit API call endpoint: 
`https://www.reddit.com/r/todayilearned/random.json`

Attempt to grab JSON data:
`var grabData = loadJSON("https://www.reddit.com/r/todayilearned/random.json");`
Fails with a warning of `Too Many Requests`.

In workshop, hacked our `loadJSON()` (doing ugly 's magic due to how getURL works):
`var grabData = loadJSON("https://www.reddit.com/r/todayilearned/random.json' --user-agent 'blahblahblah");`

With new update:
`var grabData = loadJSON("https://www.reddit.com/r/todayilearned/random.json', 'blah');`
= Success, returns JSON data since (for now) no user-agent is sending 'blah'

Tiny feature that may be more necessary if other APIs start clamping down who/what pings them. Looked through CURL options and this is the only one that I've run into needing. I want to make API calls as easy as possible, since that's a huge strength of basil.js

trych:
Ok, gave your Reddit example a try and I agree, this is a helpful extension.

I can't really test this on Windows right now, but I assume you got it to work? If that's the case feel free to merge.

> FYI – downloading (images) remains non-existant for Windows

Yes, that one should be high on our priority list.

trych (approved):

ffd8:
FYI – just ran into the user-agent issue again after finding and demo'ing the nice and simple [duckduckgo API](https://duckduckgo.com/api) – after 4 pings it stops working.. plug in a user-agent and ping away. 

And yes yes to download being high on list.. hopefully can find windows folks to help with the VBS for that. Will keep searching for examples...

trych:
Yes, as I said above, if the user-agent thing works, please go ahead and merge. The reddit example demonstrated perfectly why it is helpful.

ff6347:
Do you know this script? http://rorohiko.blogspot.com/2013/01/geturlsjsx.html It uses a different way to access data over http.

ff6347:
hm only does http as far as I can see (not https).

PR #347, opened 2020-03-05
pass collections into random()

ffd8:
- found technique to convert collections (object/arrays) into arrays to grab a random item from (as mentioned in #345 )


trych (changes requested):
This currently gives an error when the user passes an `aMin` that has no length property, like a doc or something.

trych:
I re-structured the function a bit and added better checking of the given arguments. If you approve, feel free to merge.

PR #350, opened 2020-03-10
Print correct 2.0 pre-release version when running scripts

ffd8:
Long overdue update to basil.js version that gets printed when running scripts (could be confusing when using this version versus another). package.json can be updated later once really ready.

### DIFF
--- a/basil.js
+++ b/basil.js
@@ -4958,7 +4958,7 @@ pub.SCRIPTNAME = scriptName;
  * @subcat   Constants
  * @property VERSION {String}
  */
-pub.VERSION = "1.1.0";
+pub.VERSION = "2.0.0-PRE";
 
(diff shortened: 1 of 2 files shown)


ffd8:
Long overdue detail to spit out the 2.0 dev version in console when running scripts. 
Used `2.0.0-PRE`, but open to suggestions.

trych:
Hm. Maybe `2.0.0-alpha`?

Or probably more appropriately `2.0.0-beta`?

Seems a bit more standard and therefore understandable than `PRE`, I think.

ffd8:
or simply `2.0.0-develop`?

ff6347:
+1 for `-beta` because it is open for the public. https://semver.org/

PR #353, opened 2020-03-22
docs(live preview): Adds live preview script

ff6347:
changes on the jsdoc comments can be previewed in the default documentationjs layout on http://localhost:3000

Still a bit flaky but something to address issue https://github.com/basiljs/basiljs.github.io/issues/165

close https://github.com/basiljs/basiljs.github.io/issues/165

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .DS_Store
 basil.sublime-workspace
 *.idlk
+tmp/*
 
 # Test files
 test/download/data/
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,11 +4,1187 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.8.3",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.8.3.tgz",
+      "integrity": "sha512-a9gxpmdXtZEInkCSHUJDLHZVBgb1QS0jhss4cPP93EW7s+uC5bikET2twEF3KV+7rDblJcmNvTR7VJejqd2C2g==",
+      "requires": {
+        "@babel/highlight": "^7.8.3"
+      }
+    },
+    "@babel/compat-data": {
+      "version": "7.9.0",
+      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.9.0.tgz",
+      "integrity": "sha512-zeFQrr+284Ekvd9e7KAX954LkapWiOmQtsfHirhxqfdlX6MEC32iRE+pqUGlYIBchdevaCwvzxWGSy/YBNI85g==",
+      "requires": {
+        "browserslist": "^4.9.1",
+        "invariant": "^2.2.4",
+        "semver": "^5.5.0"
+      }
+    },
+    "@babel/core": {
+      "version": "7.9.0",
+      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.9.0.tgz",
+      "integrity": "sha512-kWc7L0fw1xwvI0zi8OKVBuxRVefwGOrKSQMvrQ3dW+bIIavBY3/NpXmpjMy7bQnLgwgzWQZ8TlM57YHpHNHz4w==",
+      "requires": {
+        "@babel/code-frame": "^7.8.3",
+        "@babel/generator": "^7.9.0",
+        "@babel/helper-module-transforms": "^7.9.0",
+        "@babel/helpers": "^7.9.0",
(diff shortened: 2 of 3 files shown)


ffd8:
Hmm what am I doing wrong? Switched to this repo, ran `npm install`, followed by `npm run watch` – but [http://localhost:3000/](http://localhost:3000/) isn't running anything.

ff6347:
Ah. Sorry. It is `npm run docs` which also runs the docs thing.
It takes longer then the watching alone. Should not be the default

trych:
What is wrong with that branch? I got a error when I pulled from GH:

```
le-130-47-23:basil_2.0 trych$ git pull
remote: Enumerating objects: 42, done.                                                                                                                                                                    
remote: Counting objects: 100% (39/39), done.                                                                                                                                                             
remote: Compressing objects: 100% (11/11), done.                                                                                                                                                          
remote: Total 24 (delta 17), reused 20 (delta 13), pack-reused 0                                                                                                                                          
Unpacking objects: 100% (24/24), 63.73 KiB | 701.00 KiB/s, done.
From https://github.com/basiljs/basil.js
   5b60f3d..bb7c6d0  dev-type          -> origin/dev-type
error: cannot lock ref 'refs/remotes/origin/docs/docs-preview': 'refs/remotes/origin/docs' exists; cannot create 'refs/remotes/origin/docs/docs-preview'
 ! [new branch]      docs/docs-preview -> origin/docs/docs-preview  (unable to update local ref)
```

Is the slash in the branch name causing issues?

ff6347:
> What is wrong with that branch? 
> Is the slash in the branch name causing issues?

Don't know. I work with the slashes all day. Never had an issue with that.


did you
```bash
git fetch origin
git checkout -b docs/docs-preview origin/docs/docs-preview
```

ffd8:
Woohoo this is perfect! Immediately upon running `npm run docs` – the browser popped open to localhost:3000 with an error 'file not found' but about 5-10 seconds later, the docs appeared (I guess while it was building it the first time..)

This is really helpful for testing out composing of the docs and examples. 🙏🙏

I see the page doesn't auto-reload upon changes, but no problem, since luckily it makes adds a # with ref name to the URL, so I can simply reload the page and see the updates.

Not sure what else to test... would be great to merge so we can already use this on other dev branches.

(@trych – I could load it via the Github Desktop app without issue.. maybe worth trying there too?)

ff6347:
> I see the page doesn't auto-reload upon changes, but no problem, since luckily it makes adds a # to the URL, so I can simply reload the page and see the updates.

It should autoreload but it is really slow.

If you like it and it works for @trych we can merge and improve on this later.

ffd8:
> It should autoreload but it is really slow.

...waaaaait for it..... **reloads** - yup, it works.. wasn't being patient enough. 
No need to improve upon, just happy it's there.

trych:
Not too much time to look into this, but fine for me, so feel free to merge.

ffd8:
@fabianmoronzirfas could you please update a few lines to the [wiki#api-documentation](https://github.com/basiljs/basil.js/wiki#api-documentation)? 

And on the [Run:](https://github.com/basiljs/basil.js/wiki#development-environment-setup-os-x) instructions, should it instead suggest installing with `npm ci` instead of `npm install`, since we're including the package-lock.json file in the repo? I was running into [this issue](https://stackoverflow.com/a/56254478/10885535).

trych:
> did you
> git fetch origin

Yes, I did, even with just fetching it showed me the error and prevented me from pulling even into other branches. Now that I deleted the branch it is working again.

Not sure what exactly was the issue here, but the error message seems to suggest it is a naming issue. Could we maybe have all branch names without slashes in the future, just to be sure and avoid these problems. Thanks!

ff6347:
> Not sure what exactly was the issue here, but the error message seems to suggest it is a naming issue. Could we maybe have all branch names without slashes in the future, just to be sure and avoid these problems. Thanks!

I think it is not about the naming with slash. I work like this all the time, everyday and [I'm not the only one](https://stackoverflow.com/a/6065944/1770432). More likely that there are some conflicts with already existing branches in your local repo.

We have lots of dead branches here on the remote and I guess your local repo will have also some dead ones?

I suggest deleting merged branches by running

```bash
git branch -d BRANCH_NAME
```

If the branch is not fully merged git will warn.

Also running a 

```bash
git gc --prune=now
```

Is recommended.

trych:
> I guess your local repo will have also some dead ones?

No, I delete unused branches as soon as I don't need them anymore and have only the ones that I am currently working on/that I need to check out for reviewing. There were no naming conflicts.

Whatever, doesn't matter for now, with the deleting of the branch the problem solved itself.

PR #358, opened 2020-03-29
Dev fixmargins

ffd8:
As getter, returned an `undefined` due to lack of desired `pageNumber` with zero arguments.


trych:
Good catch. Cleaned the function up a bit more and it now allows for providing a page name or page object as parameter for the target page.

If you are okay with the changes, feel free to merge.

Issue #337, opened 2019-10-23
basil.js Plugin System


trych:
We have been talking about this in the past, I have been toying around with the idea for a while and been trying out some things.

I think we should include a simple plugin system in basil.js.

I just implemented the basics for such a system in the [plugins branch](https://github.com/basiljs/basil.js/tree/plugins) that I just uploaded and even in this basic implementation the system is already quite powerful.

The current implementation works like this: The user can place a `plugins` folder next to the `basil.js` file. Within this plugins folder there can be subfolders, each of them is considered a plugin. If basil.js finds a `index.jsx` file in such a plugin folder, the code within will be executed. All this happens before the main script flow of the user script is started. This allows to inject new functions into the global basil namespace (so, basically to add new custom basil functions) or to overwrite existing basil functions. (Later we could also add the possibility of "hooks" that trigger at specific events like document creation or after the main script flow finishes etc.)

To allow for easy disabling of plugins, a folder can be renamed to start with an underscore (`_examplePlugin`), in this case the plugin will be ignored.

Additionally there is the possibility to install "local" plugins on a per script basis, by placing a `plugins` folder next to the script file of the user. These plugins will be registered the same way, but they only refer to the script file they are placed next to (also this local `plugins` folder could further come with a list of global plugins that should be ignored to be able to further fine tune the plugin structure of each single script. This is not implemented yet in the branch at its current state). Due to the nature of ExtendScript, the local plugin folder currently works only when the script is run via Sublime Text or the InDesign Scripting panel (as I have found no way to retrieve the user script file when run from ESTK or Visual Studio Code).

Just to give you a quick and very simple example what is possible and how a plugin code could look like: Let's write a script that creates a new function `bananaAlert()` that alerts a 🍌-emoji. Let's further alter basil's `println()` function in a way that it also adds some bananas to its output. A plugin like this would simply be a `index.jsx` file that is placed into a folder `banana` next to the `basil.js` file. The `index.js` file would contain code like this:

```js
pub.bananaAlert = function() {
  alert("🍌");
}

pub.println = function() {
  // just copying basil's native println() function and tweaking it slightly
  var msg = "🍌🍌🍌 " + Array.prototype.slice.call(arguments).join(" 🍌🍌🍌 ") + " 🍌🍌🍌";
  $.writeln(msg);
  if (progressPanel)
    progressPanel.writeMessage(msg + "\n");
};
```

And then a user script like this

```js
// @include ~/Documents/basiljs/basil.js;

[...]

Issue #249, opened 2018-02-19
What to do about array methods? (forEach, map, filter)


trych:
I am slightly confused about the existing array methods in basil.

There are the methods `Array.map()` and `Array.filter()`.
Then there is the method `forEach`, which seems to do the same as JavaScript's `Array.forEach()`, but it has a different syntax `forEach(array, callback)`, so is *not* an Array method, but an independent function.

Then there are some methods that exist in JavaScript, but are missing in basil (`reduce`, `every`, `some`, `find` – any more that I am missing?).

What should we do about this? I see these options:

1. Make them more consistent and make forEach a true array method.
2. Make them more consistent and add missing ES6 array methods (reduce, every etc.).
3. Take them all out for basil.js 2.0 and let users use a more dedicated lib for such functionality (underscore.js?).

@basiljs Which of these options do you guys prefer?

fabianmoronzirfas:
1. Yes
2. We Could. See https://github.com/ExtendScript/extendscript-es6-shim and https://github.com/ExtendScript/extendscript-es5-shim
3. As the most of the audience for Basil is beginners and intermediates in classrooms I think It would be a nice helper to enable them to use modern JS functions you find all over the net in Basil without being forced to find the polyfills they need. On the other hand it creates a lot more code to watch over. :-)

trych:
Ok, so in the first step, we could just make `forEach()` consistent. Then in a second step, we could add the missing methods. Which ones are that?

`reduce`, `every`, `some`, `find`? Any more?

b-g:
1. +1 consistency 
2. could we not simply include the functionality via an external lib e.g. an old `underscore` version? That way we would not have to deal with it ...
3. +1 Would be nicer to have it shipped and included out of the box with basil 2.0

fabianmoronzirfas:
Did you take a look at  the shim repos? We could include them into the build process by installing them via npm (es5 shim is already on npm https://www.npmjs.com/package/extendscript-es5-shim) and I can publish the es6 as well. These are tested and work out of the box. No including for the enduser. Just something like 

 main.js

    //@include '../node_modules...' and so on

Or if the relativ path does not work we create some pre build process that copies the shim to the src folder.

trych:
Is this just for the few array methods? To be honest, I would just implement them via hard coding ("hardcopying") over the polyfills and be done with it. Or am I missing something here?

fabianmoronzirfas:
Going old school? :(

trych:
[...]

PR #362, opened 2020-03-30
adding indexOf() prototype for arrays

ffd8:
- added `indexOf()` for searching arrays, an often needed function that's been missing from InDesign's JS engine
- seems part of the `duplicate()` function got updated on the basil.js compiled file.. @trych - maybe your merge didn't include the most up to date compiled file??

### DIFF
--- a/basil.js
+++ b/basil.js
@@ -2805,6 +2805,15 @@ pub.matchAll = function(aString, aRegExp) {
   return results.length > 0 ? results : null;
 };
 
+Array.prototype.indexOf = function ( item ) {
+  var index = 0, length = this.length;
+  for ( ; index < length; index++ ) {
+    if ( this[index] === item )
+      return index;
+  }
+  return -1;
+};
+
 // ----------------------------------------
 // src/includes/document.js
 // ----------------------------------------
@@ -3765,7 +3774,7 @@ pub.duplicate = function(item) {
   } else if(item instanceof Page) {
 
     var newPage = item.duplicate(LocationOptions.AFTER, pub.page());
-    getAndUpdatePage(newPage, "duplicate");
+    return getAndUpdatePage(newPage, "duplicate");
 
   } else {
     error("Please provide a valid Page or PageItem as parameter for duplicate().");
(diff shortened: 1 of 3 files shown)


ffd8:
Pretty minor PR, just something I kept realizing was missing during workshops and finally add.

```
println((['bird', 'cow', 'dog'].indexOf('dog')))
println(([0, 1, 2, 3, 4, 5].indexOf(3)))
```

trych:
> added indexOf() for searching arrays, an often needed function that's been missing from InDesign's JS engine

I seem to remember that we decided against including such "ES5" Javascript features by default, as the syntax would also not be very basil-like. I'll try to find the thread about that.

> @trych - maybe your merge didn't include the most up to date compiled file??

That's possible, will have a look.

ffd8:
Hmm had missed #332 on `contains()`, which probably solves most of those situations.. 
nevertheless, it would be helpful for getting actual index, not just true/false. Of course it's a can of worms for non ES3 functions.. haha but what about including just the useful ones?? 🤠

trych:
Yep, found it, we decided in #249, to take out and not include any array methods, as it would be hard to draw a line which ones to include and they are not very basil-like, so there is no real reason to ship these polyfills with basil (unless we turn them into a proper basil method).

So, I would close this PR.

You could easily prepare a plugin with any array methods you need. Will try to get the plugin architecture included in the develop branch soon.

trych:
@ffd8 Yep, I seem to have messed up the compiling of the `duplicate()` function. Will prepare another PR for this and close this one, see my comments above.

ffd8:
Right right.. seems like best approach is the [plugins](#337 ) route – and bringing scripts like the [extendscript-es6-shim](https://github.com/ExtendScript/extendscript-es6-shim) and es5-shim into that eco-system.

trych:
Yes, we could collect all the ES5 methods in one plugin, or have separate plugins for each method, just as it suits our needs. Will try to check out the status of my plugins branch and see if it is closed to being finished, then I could hopefully prepare a PR soon.

ffd8:
Really curious what you have cooking there. I'd vote for larger plugins that solve entire issues– like this whole ES5 or ES6 as a plugin to load, rather than too many modules that aren't a simple 'download, drag + drop' for usage.

trych:
> I'd vote for larger plugins that solve entire issues– like this whole ES5 or ES6

Well, there is no real need for voting, as everyone can just create the plugin that they wish for. So, if you prefer larger plugins, go for it and create them this way. :)

PR #368, opened 2020-12-30
adding pasteInto()

ffd8:
- student duo tried to export text as a png, then import to create a clipping mask. this is akin to the GUI right-click 'pasteInto', which opens gates for wild clipping masks and within clipping masks inception!

### DIFF
--- a/basil.js
+++ b/basil.js
@@ -2663,7 +2663,7 @@ var isString = pub.isString = function(str) {
 };
 
 /**
- * @summary Checks wether a var is an InDesign text object.
+ * @summary Checks whether a var is an InDesign text object.
  * @description Checks whether a var is an InDesign text object, returns `true` if this is the case.
  * NB: a InDesign text frame will return `false` as it is just a container holding text. So you could say that `isText()` refers to all the things inside a text frame.
  *
--- a/basil.js
+++ b/basil.js
@@ -4010,6 +4010,65 @@ pub.objectStyle = function(itemOrName, props) {
 };
 
 /**
+ * @summary Paste one page item into another.
+ * @description Paste the source page item into the destination item, which then acts like a clipping mask. Optionally, set to a variable, to immediately access the new page item inside your destination.
+ *
+ * @cat     Document
+ * @subcat  Page Items
+ * @method  pasteInto
+ *
+ * @param   {PageItem} source The page item to copy.
+ * @param   {PageItem} destination The page item to paste the source into.
+ * @return  {Object} The new page item inside the destination.
+ *
+ * @example <caption>Use ellipse as clipping mask for textframe.</caption>
+ * noStroke();
+ * textSize(50);
+ *
+ * fill(0);
+ * var myText = text(LOREM, 0, 0, width, height);
+ *
+ * noFill();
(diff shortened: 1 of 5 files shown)


ffd8:
this is a minor feature – but it would help to have a quick look over incase anything is wrong..

ffd8:
small update – tried integrating `isText()` function on the newly masked pageItem.. but realized it doesn't catch it.. maybe because it's a text within the pageitem... needing to be reached with `destination.pageItems[0].hasOwnProperty('texts')`. Maybe this type of hasproptery check should find its way up the stream to the `isText()` function itself.. or just leave alone?

trych:
Hey @ffd8,
sorry for the lack of replies, I am not very active here currently as I am really occupied with other stuff.

Is there a way to "paste into" without using the actual `copy()` and `paste()` methods? I seem to recall you could avoid those entirely and that they are not supported on InDesign Server. If not, let's just go ahead and use them.

Two observations:
- Why does the text get returned when I paste the text frame into an object? I get that users most likely want to change the text eventually, but it just feels really inconsistent. What if a user actually wants to target the text frame? I think it would be preferrable if it's always the pasted object that is returned and then in this case, users have to access the `.texts[0]` property themselves.
- as it is currently implemented, this method resets the user's selection. The method should take care to reset the selection to whatever it was before.

trych (changes requested):
See my comment above. I am too dumb to use GitHub's review feature properly lol

ffd8:
@trych – noo problemo, this project jumps between front/back burners, so i totally understand. 
nevertheless, thanks for having a quick look!

I also don't really get the review feature – so just plain old comments are also preferred.

To my knowledge, `pasteInto` totally depends on the clipboard, [Jongware on pasteInto](http://jongware.mit.edu/idcs6js/pc_Application.html#pasteInto) notes: 
> Pastes data from the clipboard into the selected object ...

I originally found the solution for what to do in [this StackOverflow](https://stackoverflow.com/a/45918912/10885535).

Re: returned object when text is involved.. yeah, wanted to return whichever type of object was being fed into it (for ease with students.. feed text into mask, when set as var = your text now masked). Figuring out what to return was odd.. the same paths seem to be the same item within the mask after `pasteInto()`:
- `myMask.textFrames[0]` » expected text
- `myMask.pageItems[0].texts[0]` » expected text
- `myMask.pageItems[0].textFrames[0]` » null

I went with the later, because when feeding any other pageItem into the mask, you want the `mask.pageItems[0]` back.. so I figured that the `.pageItems[0].texts[0]` was preferable (to still work with pageItem) – but happy to use either.. just important that the actual text(Frame) gets returned. Is there actually a difference or to InDesign, it's the same object?? Happy to change it to option using `textFrames[0]`.

Re: selection, yup, you're totally right, the selection should be temp stored and returned when done! aaand changed.

ffd8:
Pulling back to front-burner for a workshop and want to include this super fun feature. Using `var maskedItem = pasteInto()` returns whatever new object has the pasted into item – if its text, get the text if shapes, the shapes.. I think it's the most intuitive for students. Surely we'll find some special cases where it breaks, but implementing to give a test run.

PR #369, opened 2022-02-18
Add improvements to b.download()

romaincazier:
Hi,

For a project I worked on, I faced issues trying to download a qrcode using Google's QR code generator:

[https://chart.googleapis.com/chart?chs=500x500&cht=qr&chl=https://basiljs.ch](https://chart.googleapis.com/chart?chs=500x500&cht=qr&chl=https://basiljs.ch)

Here are two small commits to the b.download() method, which:

- make sure all spaces in the project path are properly escaped (and not just the first one)
- avoid the query string to be stripped by the shell (because of "&") by surrounding the url with quotes

Hope it helps !

### DIFF
--- a/basil.js
+++ b/basil.js
@@ -1,4 +1,4 @@
-/* Basil.js v1.1.0 2016.11.11-15:33:03 */
+
 /*
   ..-  --.- ..- -.... -..-- .-..-. -.-..---.-.-....--.-- -....-.... -..-- .-.-..-.-.... .- .--
 
@@ -57,7 +57,7 @@
    * @property VERSION {String}
    * @cat Environment
    */
-  pub.VERSION = "1.0.10";
+  pub.VERSION = "1.1.0";
 
 /**
  * Used with b.units() to set the coordinate system to points.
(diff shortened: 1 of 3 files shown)


ff6347 (commented):
Awesome thank you. Can you also:

- maybe write a test for that change 
- run all tests and tell us the output
- run the build so the bundle is updated

romaincazier:
Hi,

I've never written tests so I hope I got it right.

The new tests I wrote for "./test/download/download.jsx" all pass, and after bundling (the tests are running with the bundled lib so I had to update it first) there seems to be no issues with all the other tests as they all passed as well.

Do you need me to give a dump of the outputs ?

ffd8:
Hey Romain! (Ted here)

Thaaanks for the PR, hooowever - this is v1.0, and we're much further along in our someday-to-release 2.0 (download also got some updates there) – check the [develop branch](https://github.com/basiljs/basil.js/tree/develop) and sorry it's not yet more clear that's where the dev/progress is taking place (we have lots of cleaning up to do with the other branches)...

romaincazier:
Hey Ted! Hope you're good!

Yeah when starting the project I first tried to check if there was a more up-to-date version but didn't look that much and seeing all the branches... I decided I'll stick with v1 with the docs and all.

Let me check the develop branch and apply my changes there if they're still relevant!

romaincazier:
Opened a new pull request for the `develop` branch

PR #370, opened 2022-02-19
Updates to b.download()

romaincazier:
See this [pull request](https://github.com/basiljs/basil.js/pull/369) on the master branch to get more informations.

### DIFF
--- a/basil.js
+++ b/basil.js
@@ -5117,7 +5117,7 @@ pub.shellExecute = function(cmd) {
  * @param   {String|File} [file] A relative file path in the project folder or a File instance
  */
 pub.download = function(url, file) {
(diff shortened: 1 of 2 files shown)


ff6347:
Hey @romaincazier Thank you for updating your PR and point it to v2 of Basil. Could you also add your tests to it?

https://github.com/romaincazier/basil.js/blob/3ec309817841b1733f65441b9bed994254b4dc0e/test/download/download.jsx

I'm not sure that making the test point to a google service is a good thing. Google has a history of shutting down services. So this could make the test fail at some point. Everything else looks fine. I can't test it locally though since I don't have any Adobe license at hand.

